Issue #2, opened 2023-04-12
feat: homepage tiles list (1 x large, 2 x small)


ramboz:
![Image](https://user-images.githubusercontent.com/1235810/229670353-b66a4a2d-9840-45ea-9c55-663c68473f12.png)
![Image](https://user-images.githubusercontent.com/1235810/229670354-536f3814-1e87-4c38-8b1e-89e9bd45e5d6.png)

## Acceptance criteria
- 1st tile of the series is larger than the others
- clicking the image on the small tiles navigates to the article (not for the large tile)
- clicking the category tag navigates to the category page
- tiles should be built from a page index so customer can just reference the page path in authoring
- date should be the published date
- author should be the author specified in the page metadata or fall back to the page author
- if the page has a video metadata, then add the video icon

Issue #16, opened 2023-04-25
feat: categories and color mappings


ramboz:
It seems that each category for the articles has a color mapped to it that is then used in the UI in a consistent way.

How do we want to handle this?
1. What is the existing mapping?
2. Should it be authorable?

![Image](https://user-images.githubusercontent.com/1235810/234344408-c86ec84b-1290-4cf8-9f87-f8325d34ca6d.png)
![Image](https://user-images.githubusercontent.com/1235810/234344406-c0b457ce-ae34-4c1b-95c2-92cee6c4ee1e.png)
![Image](https://user-images.githubusercontent.com/1235810/234344405-ea4d2be9-2fce-427f-bd75-763c79a03520.png)
![Image](https://user-images.githubusercontent.com/1235810/234344407-39a608b2-9f14-47ed-b781-543c61875c98.png)

See thread: https://adobe-dx-support.slack.com/archives/C04P5DFPDK9/p1682440568954249

ramboz:
The categories are now available at https://main--petplace--hlxsites.hlx.page/categories.json

PR #17, opened 2023-04-25
Issue 2 - Creating the tiles on the home page

travrob1:
How are the test urls used 

Fixed https://github.com/hlxsites/petplace/issues/2

Test URLs:
- After: https://issue-2--petplace--hlxsites.hlx.page/

### DIFF
--- a/blocks/tiles/tiles.css
+++ b/blocks/tiles/tiles.css
@@ -1,0 +1,103 @@
+.hero-container,
+.carousel-container,
+.search-container{
+    display: none;
+}
+
+.tile {
+    display: flex;
+    overflow: hidden;
+    margin-bottom: 1.5rem;
+}
+
+.tile .img-container {
+    position: relative;
+    min-width: 200px;
+    height: 200px;
+    border-radius: 6px;
+}
+
+.tile .img-container img {
+    position: absolute;
+    top: 0;
+    left: 0;
+    width: 100%;
+    height: 100%;
+    object-fit: cover;
+    object-position: center center;
+    opacity: 1;
+    transition: opacity 500ms ease 0s;
+    border-radius: 6px;
+}
+
+.tile .category-link-btn {
+
+}
+
+.tile .tile-contents {
+    width: 100%;
+    padding-left: 1.5rem;
+}
+
+.tile.tile-0 {
+    display: block;
+    position: relative;
+    min-height: min-content;
+}
+
+.tile.tile-0 .img-container {
+    height: auto;
+}
+
+.tile.tile-0 .tile-contents{
+    position: absolute;
+    bottom: 0;
+    left: 0;
+    background: #5773a8!important;
+    background: rgb(70 92 134 / 60%)!important;
+    color: white;
+    padding: 1.5rem 2rem 1rem;
+    border-bottom-left-radius: 6px;
+    border-bottom-right-radius: 6px;
+    display: flex;
+    justify-content: center;
+    flex-direction: column;
+    max-width: 100%;
+}
+
+.tile.tile-0 .category-link-btn {
+    position: absolute;
+    top: 1rem;
+    left: 2rem;
+    background: #5773a8;
+    box-shadow: 0 2px 4px 0 rgb(0 0 0 / 20%);
+    display: inline;
+    border-radius: 15px;
+    font-family: Raleway,sans-serif;
+    font-weight: 800;
+    font-size: .875rem;
+    color: #fff;
+    line-height: 30px;
+    height: 30px;
+    padding-left: 14px;
+    padding-right: 14px;
+}
+
+.tile.tile-0 .tile-contents h4 {
+    font-size: calc(1.325rem + .9vw);
+    color: #fff;
+    margin: 0;
+    padding-bottom: 10px;
+    text-shadow: 2px 3px 5px rgb(0 0 0 / 50%);
+    text-overflow: ellipsis;
+    line-height: 1;
+}
+
+.tile.tile-0 .tile-contents .date-author-container {
+    color: #fff;
+    text-shadow: 1px 1px 3px rgb(0 0 0 / 70%);
+    font-size: var(--body-font-size-xs);
+    font-weight: 400;
+    line-height: 1.4;
+}
+
--- a/blocks/tiles/tiles.css
+++ b/blocks/tiles/tiles.css
@@ -4,6 +4,10 @@
     display: none;
 }
 
+.tiles-wrapper {
+    width:100%;
+}
+
 .section.tiles-container {
     background: #f2f6fc;
     display: flex;
--- a/blocks/tiles/tiles.css
+++ b/blocks/tiles/tiles.css
@@ -182,7 +182,20 @@
         flex-direction: column;
         justify-content: space-between;
         width: 41.6667%;
+    }
 
+    .small-tiles-wrapper .category-link-btn {
+        background: #5773a8;
+        box-shadow: 0 2px 4px 0 rgb(0 0 0 / 20%);
+        border-radius: 15px;
+        font-family: Raleway,sans-serif;
+        font-weight: 800;
+        font-size: .875rem;
+        color: #fff;
+        line-height: 30px;
+        height: 30px;
+        padding-left: 14px;
+        padding-right: 14px;
     }
 }
 
--- a/blocks/tiles/tiles.css
+++ b/blocks/tiles/tiles.css
@@ -10,6 +10,29 @@
     flex-direction: column;
 }
 
+.section.tiles-container h2 {
+    border-bottom: 2px solid #8f8f8f;
+    margin-bottom: 2rem;
+    padding-bottom: 6px;
+    font-size: calc(1.375rem + 1.5vw);
+    line-height: 1.1;
+    font-weight: 800;
+    color: #28061e;
+}
+
+.section.tiles-container h2::after {
+    content: "";
+    display: block;
+    border-bottom: 2px solid #ff7d5a;
+    margin-top: 8px;
+    margin-bottom: -8px;
+    max-width: 200px;
+}
+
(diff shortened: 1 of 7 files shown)


blocks/tiles/tiles.css
```diff
@@ -0,0 +1,240 @@
+
+.tiles-wrapper {
+    width:100%;
+}
+
+.section.tiles-container {
```
ramboz, line 6: I wouldn't stile the container too much. We now have several styles directly on the home page template (background, headings, etc.)

blocks/tiles/tiles.css
```diff
@@ -0,0 +1,240 @@
...
+    object-fit: cover;
+    object-position: center center;
+    opacity: 1;
+    transition: opacity 500ms ease 0s;
+    border-radius: 6px;
+}
+
+.tile .category-link-btn {
+    border: none;
+    color: #6f69a8;
```
ramboz, line 64: We should probably set the color in the JS based on the category we have (i.e. hard coded mapping for now).

blocks/tiles/tiles.css
```diff
@@ -0,0 +1,240 @@
...
+    margin-top: 5px;
+    margin-bottom: 10px;
+    padding-left: 0;
+    padding-right: 0;
+    line-height: 1.2rem;
+    display: inline-block;
+    border-radius: 15px;
+    font-family: Raleway,sans-serif;
```
ramboz, line 72:
```suggestion
    font-family: var(--heading-font-family);
```

blocks/tiles/tiles.css
```diff
@@ -0,0 +1,240 @@
...
+    font-weight: 800;
+    font-size: .875rem;
```
ramboz, line 74:
```suggestion
    font-size: var(--heading-font-size-xs);
```

blocks/tiles/tiles.css
```diff
@@ -0,0 +1,240 @@
...
+    width: 100%;
+    padding-left: 1.5rem;
+}
+
+.tile .tile-contents h4 {
+    margin-bottom: 0.5rem;
+    font-size: 1rem;
+    line-height: 1;
+    font-weight: 800;
+    color: #28061e;
```
ramboz, line 87: I would propose to move the color and weight to the global `h1, h2, h3, h4, h5, h6` styles.

blocks/tiles/tiles.css
```diff
@@ -0,0 +1,240 @@
...
+}
+
+.tile .tile-contents .date-author-container {
+    font-size: .875rem;
+    font-weight: 400;
```
ramboz, line 92: This and below styles should all be inherited from the body now

blocks/tiles/tiles.css
```diff
@@ -0,0 +1,240 @@
...
+.tile.tile-0 .img-container .img-padding {
+    width:100%;
+    padding-bottom: 75%;
+}
+
+.tile.tile-0 .tile-contents{
+    position: absolute;
+    bottom: 0;
+    left: 0;
+    background: #5773a8!important;
```
ramboz, line 118: These 2 colors would probably best be set in the JS from the article category, via hard coded mapping. You also don't need the `!important` modifier, I think

blocks/tiles/tiles.css
```diff
@@ -0,0 +1,240 @@
...
+    justify-content: center;
+    flex-direction: column;
+    max-width: 100%;
+}
+
+.tile.tile-0 .category-link-btn {
+    position: absolute;
+    top: -15px;
+    left: 2rem;
+    background: #5773a8;
```
ramboz, line 134: Same here, should probably be set in the JS

blocks/tiles/tiles.css
```diff
@@ -0,0 +1,240 @@
...
+    box-shadow: 0 2px 4px 0 rgb(0 0 0 / 20%);
+    border-radius: 15px;
+    font-family: Raleway,sans-serif;
+    font-weight: 800;
+    font-size: .875rem;
+    color: #fff;
+    line-height: 30px;
```
ramboz, line 141: I would typically round this up to the nearest "rem". 30px => 2 rem, and 14px => 1rem below
Tends to simplify the CSS and makes it a bit more consistent on the padding/spacing

blocks/tiles/tiles.css
```diff
@@ -0,0 +1,240 @@
...
+    display: none;
+}
+
+.tile.tile-0 .category-link-btn.category-link-btn-mobile {
+    display: inline;
+    top: 1rem
+}
+
+.tile.tile-0 .tile-contents h4 {
+    font-size: calc(1.325rem + .9vw);
```
ramboz, line 154: I'd remove the calc and vw modifier

blocks/tiles/tiles.css
```diff
@@ -0,0 +1,240 @@
...
+    color: #fff;
+    text-shadow: 1px 1px 3px rgb(0 0 0 / 70%);
+    font-size: var(--body-font-size-xs);
+    font-weight: 400;
+    line-height: 1.4;
+}
+
+@media screen and (min-width: 461px) {
+    .tile .tile-contents h4 {
+        font-size: calc(1.2625rem + .15vw);
```
ramboz, line 173: Same

blocks/tiles/tiles.css
```diff
@@ -0,0 +1,240 @@
...
+
+    .small-tiles-wrapper {
+        display: flex;
+        flex-direction: column;
+        justify-content: space-between;
+        width: 41.6667%;
+    }
+
+    .small-tiles-wrapper .category-link-btn {
+        background: #5773a8;
```
ramboz, line 217: Color is probably best set via JS

blocks/tiles/tiles.css
```diff
@@ -0,0 +1,240 @@
...
+        box-shadow: 0 2px 4px 0 rgb(0 0 0 / 20%);
+        border-radius: 15px;
+        font-family: Raleway,sans-serif;
```
ramboz, line 220: Font props should already be inherited from the mobile styles, I think

blocks/tiles/tiles.js
```diff
@@ -0,0 +1,110 @@
+export default async function decorate(block) {
+  // Create containing div of three tiles (one big, two small)
+  const tileContainer = document.createElement('div');
+  tileContainer.className = 'tiles-container';
```
ramboz, line 4: The`.section` will already be annotated with `.tiles-container` by Franklin as it contains a `.tiles` block, so you may have side-effects on any styles you set on this.

You could still use it as a virtual container and just do:
```js
block.innerHTML = tileContainer.innerHTML;
```
at the end

blocks/tiles/tiles.js
```diff
@@ -0,0 +1,110 @@
...
+
+  const data = await new Promise((resolve, reject) => {
```
ramboz, line 6: I'm not sure you need the wrapping Promise, you might just as well directly assign the `Promise.all` result to the `data` variable

blocks/tiles/tiles.js
```diff
@@ -0,0 +1,110 @@
...
+    const urls = [...block.children].map((row) => {
+      const path = new URL(row.firstElementChild.firstElementChild.text).pathname;
+      return `https://admin.hlx.page/index/hlxsites/petplace/main/${path}`;
```
ramboz, line 9: *nitpicking* you probably have a double slash in the resulting URL
```suggestion
      return `https://admin.hlx.page/index/hlxsites/petplace/main${path}`;
```

blocks/tiles/tiles.js
```diff
@@ -0,0 +1,110 @@
...
+    categoryLink.href = dta.path.substring(0, dta.path.lastIndexOf('/'));
+    categoryLink.innerHTML = dta.category;
+
+    const categoryLinkMobile = categoryLink.cloneNode(true);
+    categoryLinkMobile.classList.add('category-link-btn-mobile');
+
+    const title = document.createElement('a');
+    title.href = dta.path;
+    const titleHeader = document.createElement('h4');
+    titleHeader.innerHTML = dta.title.substring(0, dta.title.lastIndexOf(' - PetPlace'));
```
ramboz, line 67: We'll have to keep in mind this suffix handling (` - PetPlace`), I'm going to assume they are likely not consistent with this on all their pages. Might need a later iteration once we have more sample pages

blocks/tiles/tiles.css
```diff
@@ -0,0 +1,240 @@
+
+.tiles-wrapper {
+    width:100%;
```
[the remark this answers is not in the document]
travrob1: Very much needed as without the tiles get squished.
image

blocks/tiles/tiles.css
travrob1: I can remove the background... but i need to keep the flex and the direction

blocks/tiles/tiles.css
travrob1: I would suggest we leave this for now... since we will be addressing the colors  in
https://github.com/hlxsites/petplace/issues/16

blocks/tiles/tiles.css
travrob1: This is another color that i think we address as part of 
https://github.com/hlxsites/petplace/issues/16
Again i would suggest we leave it for now and address it as a whole later.

blocks/tiles/tiles.css
travrob1: Same as comment above ☝️

blocks/tiles/tiles.js
travrob1: Good catch.

blocks/tiles/tiles.js
travrob1: That is indeed correct.

ramboz (approved):
Looks good to me.

PR #51, opened 2023-04-28
Issue 1 - Adding responsive Carousel and Responsive Images

### LINKED ISSUE #1: feat: homepage slideshow
Image
Image
Image

Requirements:
- responsive layout
- responsive images (<440px, <769px and above)
- accessible (see: https://www.w3.org/WAI/ARIA/apg/patterns/carousel/)

### LINKED ISSUE #5: feat: automatically decorate responsive images
Petplace uses responsive images where they have different variants for each viewport extensively.

For instance, we have this in the image carousel on the homepage:

Or this for the hero on breeds index page:

Image
Image

We should automatically decorate those to single `<picture>` tag with different variants for each viewport on our end.
In the document we could have those different images just as siblings, like:

Image
Image

and then auto-decorate and combine to a single image (globally like we do icons).

We did something similar in https://github.com/hlxsites/delta/blob/main/us/en/skymiles/scripts/scripts.js#L78-L86 for Delta

travrob1:
Fix #1 
Fix #5 

Test URLs:
- After: https://issue-1--petplace--hlxsites.hlx.page


blocks/slideshow/slideshow.css
```diff
@@ -0,0 +1,111 @@
...
+}
+
+.slides-container .slide {
+    width: 100vw;
+    flex-shrink: 0;
+    position: relative;
+}
+
+.slides-container [aria-hidden="true"] {
+    display: block;
```
ramboz, line 28: From an accessibility point of view, if it is "hidden" it shouldn't be visible. It would rather be best to toggle to `aria-hidden="false"`

blocks/slideshow/slideshow.css
```diff
@@ -0,0 +1,111 @@
...
+    height: 32px;
+}
+
+.tab-bar-container > ol svg {
+    width: 20px;
+    height: 20px;
+}
+
+.control-button:hover svg {
+    color: #7f7f7f;
```
ramboz, line 68: We have `--color-gray: #8f8f8f;` in the global styles, might be close enough to use and avoid an additional color to manange

blocks/slideshow/ResumeableInterval.js
```diff
@@ -0,0 +1,26 @@
+class ResumableInterval {
```
ramboz, line 1: I'd be tempted to move the class back inside slideshow.js to avoid an additional HTTP roundtrip for just a few lines of JS

icons/circle-fill.svg
ramboz, line 1: I would modify the icon so:
1. the white part is hardcoded on the `circle` with the `fill` property for the 1st one
2. the 2nd one uses `fill="currentcolor"` instead so it inherits from the CSS color and makes styling easier
3. you remove the `style` from the file

icons/circle.svg
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Generator: Adobe Illustrator 27.5.0, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
+<svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
+	 viewBox="0 0 16 16" style="enable-background:new 0 0 16 16;" xml:space="preserve">
+<style type="text/css">
```
ramboz, line 5: Same here, remove the style and inline the color as a `fill` property

scripts/scripts.js
```diff
@@ -122,6 +130,38 @@ function loadDelayed() {
...
+    responsivePicture.prepend(srcElemBackup);
+    responsivePicture.prepend(srcElem);
+  });
+
+  return responsivePicture;
+}
+
+export function decorateResponsiveImages(container) {
+  const responsiveImage = createResponsiveImage([...container.querySelectorAll('picture')]);
+  container.innerHTML = responsiveImage.outerHTML;
```
ramboz, line 162: In my experience, using `outerHTML` is typically more expensive than just doing:
```js
container.innerHTML = '';
container.append(responsiveImage);
```

icons/chevron-left.svg
```diff
@@ -0,0 +1,3 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" fill="currentColor" class="bi bi-chevron-left" viewBox="0 0 16 16">
```
ramboz, line 1: I would typically clean up the svg file and remove `width`, `height` and `class` on the to element

blocks/slideshow/slideshow.css
travrob1: The problem here is the way the slideshow works.  Imagine a row of pictures...each picture is the length of the viewport.  If you hide one via display block, then the order of the pictures is thrown off.  I struggled a long time with this issue.  So visually the other slides are hidden...because they are off of the page.  If this is a show stopper, then i need to re engineer the slideshow another way.

blocks/slideshow/ResumeableInterval.js
travrob1: done

blocks/slideshow/slideshow.css
ramboz: Let's address that in a dedicated issue to make this fully accessible. Probably part of our 2nd Milestone when we do all the accessibility fine-tuning

blocks/slideshow/slideshow.css
```diff
@@ -0,0 +1,108 @@
...
+}
+
+.slides-container [aria-hidden="true"] {
+    display: block;
+}
+
+
+.slide .text-div {
+    position: absolute;
+    left: 0;
```
ramboz, line 34: You can use: `inset: 0 0 0 0` to combine the 4 `top/right/bottom/left` props together.

scripts/lib-franklin.js
```diff
@@ -174,7 +174,7 @@ export async function decorateIcons(element) {
   svgSprite.innerHTML += symbols;
 
   icons.forEach((span) => {
-    const iconName = Array.from(span.classList).find((c) => c.startsWith('icon-')).split('-')[1];
+    const iconName = Array.from(span.classList).find((c) => c.startsWith('icon-')).substring(5);
```
ramboz, line 177: I would suggest you also take 10mins and open a PR against https://github.com/adobe/helix-project-boilerplate/blob/main/scripts/lib-franklin.js#L183 and do it there. That way you fix it at the source, and you also get a commit there for visibility 😉

styles/styles.css
```diff
@@ -125,14 +132,19 @@ h3, h4, h5, h6 {
   margin-top: 0;
 }
 
-h1 { font-size: var(--heading-font-size-xxl) }
+h1 {
+  font-size: var(--heading-font-size-xxl);
+  color: #fff;
```
ramboz, line 137:
```suggestion
  color: var(--text-color-inverted);
```

ramboz (approved):
Looks good to me!

Just minor comments, and we can defer the a11y work to a different PR

PR #58, opened 2023-05-03
adding search styles and behavior

### LINKED ISSUE #27: feat: homepage search
Image

travrob1:
Adding styles and some behavior for the home page search

Fix #27

Test URLs:
- After: https://issue-27--petplace--hlxsites.hlx.page/


ramboz:
I would suggest 2 minor adjustments:
1. On mobile, the rounder border on the original site didn't touch the page borders. While I wouldn't necessarily do it exactly how they had it, I would still mimic this slightly. Maybe something like `width: 80%; padding-left: 0; padding-right: 0;` on the heading for mobile breakpoint might do it
2. The font-size on tablet is bigger on our end as it was on the initial site. While I wouldn't follow their `vw` dynamic sizing, we can probably set a static intermediate font-size for the tablet breakpoint

| | Franklin | Original |
|-|-|-|
|Mobile|Screenshot 2023-05-03 at 2 15 41 PM|Screenshot 2023-05-03 at 2 15 43 PM|
|Tablet|Screenshot 2023-05-03 at 2 16 25 PM|Screenshot 2023-05-03 at 2 16 29 PM|

blocks/search/search.css
```diff
@@ -0,0 +1,97 @@
+.section.search-container {
```
ramboz, line 1: *nitpicking* you don't really need the double selector `.section.search-container`, `.search-container` should be just fine

blocks/search/search.css
```diff
@@ -0,0 +1,97 @@
...
+    bottom: 60px;
+}
+
+.section.search-container h2::before,
+.section.search-container h2::after {
+    content: " ";
+    position: absolute;
+    width: 60px;
+    height: 36px;
+    background-image: url("data:image/svg+xml;base64,PHN2ZyB3aWR0aD0iMTIxIiBoZWlnaHQ9IjM2IiB2aWV3Qm94PSIwIDAgMTIxIDM2IiBmaWxsPSJub25lIiB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciPjxwYXRoIGQ9Ik05Mi44MDgyIDE5LjQ3OTRDODYuNzM3NSA3Ljg5MDQgNzQuNTk1OSAtMS4yNDgwNGUtMDUgNjAuNDU4NCAtMS4wMTE1MWUtMDVDNDYuNDA0MSAtNy43NjM3M2UtMDYgMzQuMTc5NCA3Ljg5MDQxIDI4LjEwODYgMTkuNDc5NUMyMi43MDMxIDI5Ljc1MzQgMTEuNzI1OCAzNiA2LjU3ODA3ZS0wNiAzNkw2MC41IDM2TDEyMSAzNkMxMDkuMjc0IDM2IDk4LjIxMzcgMjkuNzUzNCA5Mi44MDgyIDE5LjQ3OTRaIiBmaWxsPSJ3aGl0ZSIvPjwvc3ZnPg==");
```
ramboz, line 24: I would propose to extract this to its own svg file. You can use the browser to open the image in a new tab and then copy the SVG markup from there.

blocks/search/search.css
```diff
@@ -0,0 +1,97 @@
...
+    padding: 40px 30px 0;
+    margin: 0 ;
+    bottom: 60px;
+}
+
+.section.search-container h2::before,
+.section.search-container h2::after {
+    content: " ";
+    position: absolute;
+    width: 60px;
```
ramboz, line 22: This 60px that is used for the width and then later to offset left and right could be extracted to a CSS variable so it's easier to update at once

blocks/search/search.css
```diff
@@ -0,0 +1,97 @@
...
+    height: 36px;
+    background-image: url("data:image/svg+xml;base64,PHN2ZyB3aWR0aD0iMTIxIiBoZWlnaHQ9IjM2IiB2aWV3Qm94PSIwIDAgMTIxIDM2IiBmaWxsPSJub25lIiB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciPjxwYXRoIGQ9Ik05Mi44MDgyIDE5LjQ3OTRDODYuNzM3NSA3Ljg5MDQgNzQuNTk1OSAtMS4yNDgwNGUtMDUgNjAuNDU4NCAtMS4wMTE1MWUtMDVDNDYuNDA0MSAtNy43NjM3M2UtMDYgMzQuMTc5NCA3Ljg5MDQxIDI4LjEwODYgMTkuNDc5NUMyMi43MDMxIDI5Ljc1MzQgMTEuNzI1OCAzNiA2LjU3ODA3ZS0wNiAzNkw2MC41IDM2TDEyMSAzNkMxMDkuMjc0IDM2IDk4LjIxMzcgMjkuNzUzNCA5Mi44MDgyIDE5LjQ3OTRaIiBmaWxsPSJ3aGl0ZSIvPjwvc3ZnPg==");
+    top: 0;
+    z-index: 10;
+    left: -60px
+}
+
+.section.search-container h2::after {
+    display: block !important;
```
ramboz, line 31: We tend to stay away from `!important`, and you can probably achieve the same via more CSS specificity:
`.home-page .search-container h2::after`

blocks/search/search.css
```diff
@@ -0,0 +1,97 @@
...
+    color: var(--background-color);
+    text-transform: uppercase;
+    border: 1px solid hsl(0deg 0% 100% / 78%);
+    font-size: 1.5rem;
+    font-weight: 800;
+    font-family: var(--heading-font-family);
+
+}
+
+@media (min-width: 461px) {
```
ramboz, line 81: I would stick to `600px`. This will slightly change the layout on the 460~600 viewports, but it avoids having an additional media query to handle at the project level. We probably should have done this as well for the tiles one

blocks/search/search.js
```diff
@@ -0,0 +1,33 @@
+export default async function decorate(block) {
+  let searchPlaceholder = 'Search for Posts...';
```
ramboz, line 2: I would get those from the initial block that is passed down to you:
```js
searchPlaceholder = block.firstElementChild.children[0].textContent || 'Search…';
searchButtonText = block.firstElementChild.children[1].textContent || 'Submit';
```
or similar.

That way the actual content gets changed in the docx directly and you don't need to handle this in the code

blocks/search/search.js
```diff
@@ -0,0 +1,33 @@
...
+    searchButtonText = 'Submit';
+  }
+
+  const form = document.createElement('form');
+  form.className = 'search-box-wrapper';
+  form.action = '/search';
+
+  const input = document.createElement('input');
+  input.className = 'search-input';
+  input.type = 'text';
```
ramboz, line 15: you could use `type="search"` so you inherit from the native browser UI and accessibility for it

blocks/search/search.js
```diff
@@ -0,0 +1,33 @@
...
+  input.name = 'query';
+  input.placeholder = searchPlaceholder;
+
+  form.append(input);
+
+  const button = document.createElement('button');
+  button.type = 'submit';
+  button.className = 'search-button';
+  button.innerHTML = searchButtonText;
```
ramboz, line 24:
```suggestion
  button.textContent = searchButtonText;
```

blocks/search/search.js
```diff
@@ -0,0 +1,33 @@
...
+
+  form.append(button);
+  block.textContent = '';
```
ramboz, line 27:
```suggestion
  block.innerHTML = '';
```

blocks/search/search.css
travrob1: nope .... it aint ;)

blocks/search/search.css
travrob1: ended up needing to offset the with and left by one to avoid a gap that can appear when resized.

blocks/search/search.js
travrob1: good call.

travrob1:
The text size on tablet you mentioned... i couldn't see the difference you mentioned.

ramboz (approved):

PR #60, opened 2023-05-04
feat: table of contents block

### LINKED ISSUE #3: feat: table of contents block
Image

## Acceptance Criteria
- Automatically build from the `<h2>` elements that follow it in the section
- Automatically numbered
- Clicking an entry in the ToC scrolls to the corresponding header and moves the focus to it

## Sample pages
- https://www.petplace.com/article/cats/pet-care/cat-drinking-fountains/

varunmitra:

Fix #3 

Test URLs:
- Before: https://main--petplace--hlxsites.hlx.page/article/cats/pet-care/cat-drinking-fountains
- After: https://issue-3--petplace--hlxsites.hlx.page/article/cats/pet-care/cat-drinking-fountains


styles/styles.css
```diff
@@ -456,3 +456,59 @@ a.sr-focusable:focus {
 .article-page main li {
   margin-bottom: .25rem;
 }
+
+@media (min-width: 600px) {
+  main .section.toc > div > h2 {  
+    font-size: calc(1.375rem + 1.5vw);
```
ramboz, line 462: I would stick to static font-sizes, the fluid `vw` increments essentially require a re-layout any time we resize the window, and the different in font size you get from it isn't really perceptible.

styles/styles.css
```diff
@@ -456,3 +456,59 @@ a.sr-focusable:focus {
...
+    line-height: 1.1;  
+    margin-block-start: 0.83em;
```
ramboz, line 464: These props feel a bit like a copy-paste from the browser's style panel.
There are short-hand notations for those we could use, and the regular margin props are usually enough. or did you really need to offset the margin computation?

styles/styles.css
```diff
@@ -456,3 +456,59 @@ a.sr-focusable:focus {
...
+    margin-block-end: 0.83em;
+    margin-inline-start: 0;
+    margin-inline-end: 0;  
+  }
+  
+  main .section.toc > div > ol {
+    display: block;
+    list-style-type: decimal;
+    margin-block-start: 1em;
```
ramboz, line 473: Same here

styles/styles.css
```diff
@@ -456,3 +456,59 @@ a.sr-focusable:focus {
...
+    line-height: 1.5;
+    font-weight: 400;
+  }
+}
+
+@media (min-width: 900px) {
+  main .section.toc > div > h2 {  
+    margin-top: 1.5rem;
+    margin-bottom: 1rem;
+    font-size: calc(1.375rem + 1.5vw);
```
ramboz, line 497: These 2 are already inherited from the mobile styles

styles/styles.css
```diff
@@ -456,3 +456,59 @@ a.sr-focusable:focus {
...
+    line-height: 1.1;
+  }   
+}
+
+@media (min-width: 1200px) {
+  main .section.toc > div > h2 {  
+    margin-top: 1.5rem;
+    margin-bottom: 1rem;
+    font-size: 2.5rem;
```
ramboz, line 506: You can remote the other props on this block. They are inherited from the previous media queries, and the font weight is already set on the font definition itself

ramboz (commented):
At a high-level, I think we can automatically build the ToC based on the h2s in the document.
I don't think we need the author to manually author this.

A plain ToC block in the content will likely be faster to author than manually setting all the anchors and links. We could maybe even auto-block it completely for each article page if it has more than 3 h2s and/or more than a predefined set of characters in the text content.

ramboz (approved):
Looks good to me. Jut needs fixing the linting issue

scripts/scripts.js
```diff
@@ -95,14 +96,41 @@ function buildHeroBlock(main) {
...
+ * @param {Element} main The container element
+ */
+function buildTOCBlock(main) {
+  const tocDiv = document.createElement('div');
+  const allH2s = main.getElementsByTagName('h2');
+  const tocHeader = document.createElement('h2');
+  const tocList = document.createElement('ol');
+  tocHeader.innerText = 'Table of Contents';
+  tocDiv.appendChild(tocHeader);
+  if (allH2s.length > 1 && getMetadata('template') === 'Article Page') {
```
ramboz, line 110: I would suggest 2 small changes:
1. Move the conditional directly in the `buildAutoBlocks` function like we have for the category index
2. Use the body class `document.body.classList.contains('article-page')` as the technical name is less prone typos (as we sanitize it)

PR #62, opened 2023-05-05
add templates and dynamically load them.

### LINKED ISSUE #8: feat: implement template system for the various pages
- https://www.petplace.com/

mfrisbey:
Fix #8 

Test URLs:
- After: https://issue-8--petplace--hlxsites.hlx.page/

This PR adds the structure for templates, and provides the ability for pages to specify which template should be used. Currently there are 3 templates:

* `article`: for information pages similar to https://www.petplace.com/article/dogs/pet-care/dog-care/grooming-your-dog/the-ultimate-how-to-guide-for-grooming-a-dog-at-home/
* ` bio`: for breed pages similar to https://www.petplace.com/article/breed/bernese-mountain-dogs/
* `category`: for pages with tiles, similar to https://www.petplace.com/article/breed/

Pages can specify which template to use through their metadata, similar to:

| Metadata | |
| --------- | --------- |
| Template | article |

Specifying this template metadata will cause the page to load the `article.css` stylesheet, which will apply the general structure and styles of that template.

The css files are currently empty, but I'm working on filling them out. I wanted to get this PR merged early so that we could at least start specifying the template metadata on the pages we're focusing on.


templates/bio.css
travrob1, line 1: would bio be used for the breeds?
https://www.petplace.com/article/breed/boston-terrier/ 
for example?

mfrisbey: @travrob1 yes, that's the idea for that one.

scripts/scripts.js
```diff
@@ -12,6 +12,15 @@ import {
   loadBlocks,
   loadCSS, getMetadata,
 } from './lib-franklin.js';
+import article from '../templates/article/article.js';
+import bio from '../templates/bio/bio.js';
+import categories from '../templates/categories/categories.js';
+
+const TEMPLATES = {
```
ramboz, line 19: Let's make sure we align the names of the templates with what we define in [metadata.xslx](https://adobe.sharepoint.com/:x:/r/sites/HelixProjects/Shared%20Documents/sites/petplace/metadata.xlsx?d=wbe274adbc57d4c61a6f5f6275a24b517&csf=1&web=1&e=Ior3pI) so we get automatic decoration for those without having to explicitly specify the metadata in the article/breed page in authoring

scripts/scripts.js
```diff
@@ -72,6 +81,14 @@ function buildCategorySidebar() {
   return section;
 }
 
+/**
+ * Retrieves the name of the template as specified in the page's metadata.
+ * @returns {string} The name of the page's template, or undefined if none specified.
+ */
+function getTemplateName() {
+  return document.head.querySelector('meta[name="template"]')?.getAttribute('content');
```
ramboz, line 89: Technically, you can leverage the `getMetadata` in lib-franklin.js to achieve this

scripts/scripts.js
```diff
@@ -158,6 +175,22 @@ function buildTOCBlock(main) {
...
+ *  been loaded.
+ */
+async function buildTemplateBlock(main) {
+  const template = getTemplateName();
+  if (template) {
+    const templateLoader = TEMPLATES[template];
+    if (templateLoader) {
+      await templateLoader(main);
+    }
+  }
```
ramboz, line 191: *nitpicking* I always suggest using "guard clauses" to limit code indentation
```suggestion
  if (!template) {
    return;
  }
  const templateLoader = TEMPLATES[template];
  if (!templateLoader) {
    return;
  }
 return templateLoader(main);
}
```

scripts/scripts.js
```diff
@@ -12,6 +12,15 @@ import {
   loadBlocks,
   loadCSS, getMetadata,
 } from './lib-franklin.js';
+import article from '../templates/article/article.js';
```
ramboz, line 15: Instead of importing all the templates initially, I'd suggest moving this down into the `buildTemplateBlock` method:
```js
const default = await import('../templates/${template}/${template}.js`);
default(main);
```

ramboz (commented):
Looking good so far!

scripts/scripts.js
mfrisbey: Thanks, I've updated metadata.xslx and changed the names of the templates as needed to match.

scripts/scripts.js
mfrisbey: Done!

scripts/scripts.js
mfrisbey: I see your point. There's an argument in there for indentation vs multiple return statements (which I usually try to avoid), but I'm happy to follow the project style. I've updated the code as you suggested.

scripts/scripts.js
mfrisbey: Nice! I've run into pushback in the past regarding dynamically importing files, but if it's ok then that will really simplify the code. I've made the change.

mfrisbey:
@ramboz I've made changes based on your feedback - would you please take another quick look? Thanks!

mfrisbey:
Interestingly enough, it seems like dynamically importing the template JS may have impacted the LH score (it was 100% before the last commit). I know there are some optimizations when using `import` (i.e. the code is only imported if it's actually used) at the top level - maybe it's actually better to avoid the dynamic import?

travrob1:
@mfrisbey I have learned that if you see a surprisingly slow speed... just re-run the psi check via the checkmark above.  Your performance looks better now ;)

scripts/scripts.js
```diff
@@ -204,6 +233,10 @@ export function decorateMain(main) {
 async function loadEager(doc) {
   document.documentElement.lang = 'en';
   decorateTemplateAndTheme();
+  const template = getTemplateName();
```
ramboz, line 236: I'd be tempted to move this into the `buildTemplateBlock` so we load both js and CSS in the same place. it will also avoid fetching the template tag twice as a side effect

templates/article/article.css
```diff
@@ -0,0 +1,97 @@
+main .hero-container .hero h1 {
+  font-size: calc(1.4875rem + 2.85vw);
```
travrob1, line 2: petplace.com used dynamic font sizes based on screen size but so far we have opted to NOT follow the trend.  Julien informed me it just adds a close approximation standard size

PR #65, opened 2023-05-05
breed general attributes block and styles

### LINKED ISSUE #35: feat: breed general attributes block
Image
Image

travrob1:
Adding the general attributes block to the breeds page.
Additional adding an auto block for the author.

Fix #35

Test URLs:
- After: https://issue-35--petplace--hlxsites.hlx.page/article/breed/pembroke-welsh-corgi


icons/user.svg
travrob1, line 1: looks all black because i set currentcolor to both path and circle to be controlled with css selector

ramboz (approved):

PR #69, opened 2023-05-10
Flesh out breed template

### LINKED ISSUE #8: feat: implement template system for the various pages

mfrisbey:

Fix #8

Test URLs:
- After: https://issue-8-breed-template--petplace--hlxsites.hlx.page/

This PR has a few changes:

* Add placeholders in all unfinished blocks used by current templates.
* Move table of contents creation from main scripts.js to article template.
* Add ability for templates to control the hero block creation (the breeds template has a unique hero layout)
* Finish mobile layout for breeds template.

### DIFF
--- a/scripts/scripts.js
+++ b/scripts/scripts.js
@@ -27,10 +27,10 @@ const TEMPLATES = {
 
 /**
  * @typedef TemplateLoader
- * @property {function} buildTemplateBlock Accepts a single argument, a target element, that will add blocks
- *  specific to a given template.
- * @property {function} [buildHeroBlock] Accepts 3 arguments: a target element, the hero picture, and the
- *  hero text. The function will add blocks required for the template's hero section.
+ * @property {function} buildTemplateBlock Accepts a single argument, a target element, that will
+ *  add blocks specific to a given template.
+ * @property {function} [buildHeroBlock] Accepts 3 arguments: a target element, the hero picture,
+ *  and the hero text. The function will add blocks required for the template's hero section.
  */
 
 const LCP_BLOCKS = []; // add your LCP blocks to the list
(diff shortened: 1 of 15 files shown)


scripts/scripts.js
```diff
@@ -73,6 +92,30 @@ function buildCategorySidebar() {
...
+
+/**
+ * Retrieves the template applicable to the current page, if there is a template.
+ * @returns {TemplateLoader} Loader for providing various functionality for interacting with
+ *  a template. May be undefined if there is no valid template specified.
+ */
+function getTemplateLoader() {
+  const template = getTemplateName();
+  if (!template) {
+    return undefined;
```
ramboz, line 111: *nitpicking* I usually recommend to explicitly return `null` instead of `undefined`. This usually helps long-term to maintain the code and detect edge cases that might not be properly covered:
- `null`: the developer intentionally returned this value, so we have a valid use case
- `undefined`: the developer didn't handle this explicitly, so you likely found an edge case that isn't covered.

templates/article-page/article-page.js
```diff
@@ -5,21 +5,20 @@ import {
 function createTemplateBlock(main, blockName, gridName) {
   const gridNameValue = gridName || blockName;
   const section = document.createElement('div');
-  section.className = `article-template-autoblock article-template-grid-${gridNameValue}`;
+  section.className = `section article-template-autoblock article-template-grid-${gridNameValue}`;
```
ramboz, line 8: If you use:
```suggestion
  section.classList.add('article-template-autoblock', `article-template-grid-${gridNameValue}`);
```
you don't have to add the default class again since you are just appending to the list

ramboz (approved):
Looks good overall. Added a few minor comments

scripts/scripts.js
```diff
@@ -73,6 +92,30 @@ function buildCategorySidebar() {
   return section;
 }
 
+/**
+ * Retrieves the name of the template being used by the current page.
+ * @returns {string} Template name, or undefined if none.
+ */
+function getTemplateName() {
+  return getMetadata('template')?.toLowerCase();
```
ramboz, line 100: You can use the `toClassName` from `lib-franklin.js` to get a standardized name from the string

scripts/scripts.js
```diff
@@ -184,14 +173,22 @@ async function buildTemplateBlock(main) {
...
   }
+  const templateLoader = getTemplateLoader();
+  if (!templateLoader) {
+    return;
+  }
   try {
-    await loadCSS(`/templates/${template}/${template}.css`);
-    const templateLoader = await import(`../templates/${template}/${template}.js`);
-    await templateLoader.default(main);
+    templateLoader.buildTemplateBlock(main);
```
ramboz, line 181: You're somehow missing an `await import` for the template JS files around here

PR #71, opened 2023-05-11
Create layout for tablet resolution in articles

### LINKED ISSUE #8: feat: implement template system for the various pages

mfrisbey:
Fix #8 

Test URLs:
- After: https://issue-8-article-tab--petplace--hlxsites.hlx.page/

Some minor layout changes to accommodate a footer below the article. The css in the template will use its grid to reposition all elements into the column/row layout of the tablet view.


ramboz (commented):
Looks good. I think the perf issues are regressions on main. Let me fix that first and we'll see if we can get the score back in the in the PR

PR #75, opened 2023-05-11
fix: address recent performance regressions

ramboz:
1. Modify the template loading logic so it loads both JS & CSS in parallel to speed things up
2. Renamed navigation block to breadcrumb for the article paged and injected into the hero container
3. Fix general attributes decoration for the breed pages
4. Replace all `outerHTML` usage by the faster `append`

Test URLs:
- After:
  - https://perf-improvements--petplace--hlxsites.hlx.page/
  - https://perf-improvements--petplace--hlxsites.hlx.page/article/cats/pet-health/cat-health/cat-diet-nutrition/the-ultimate-guide-to-what-cats-cant-eat
  - https://perf-improvements--petplace--hlxsites.hlx.page/article/breed/pembroke-welsh-corgi


scripts/scripts.js
```diff
@@ -131,16 +92,16 @@ async function buildHeroBlock(main) {
...
-
-    let templateBuilt = false;
-    const templateLoader = await getTemplateLoader();
-    if (templateLoader && templateLoader.buildHeroBlock) {
-      templateBuilt = true;
-      templateLoader.buildHeroBlock(section, picture, h1);
-    }
-    if (!templateBuilt) {
+    if (bodyClass.includes('breed-page')) {
+      section.append(buildBlock('hero', { elems: [picture] }));
```
mfrisbey, line 96: I added the `buildHeroBlock` method so that the template js files would be where logic like this lives. Do you want to keep this pattern of `if` statements in the main script instead?

scripts/scripts.js
```diff
@@ -131,16 +92,16 @@ async function buildHeroBlock(main) {
...
+    } else {
       section.append(buildBlock('hero', { elems: [picture, h1] }));
     }
+    if (bodyClass.includes('article-page')) {
+      const breadcrumb = document.querySelector('.breadcrumb');
+      breadcrumb.parentElement.remove();
+      section.append(breadcrumb);
+    }
```
mfrisbey, line 104: Similar comment here - this is something that could potentially live in the article template's `buildHeroBlock` method if we wanted to keep it out of the main scripts file.

scripts/scripts.js
```diff
@@ -231,6 +199,7 @@ async function loadEager(doc) {
   decorateTemplateAndTheme();
   const main = doc.querySelector('main');
   if (main) {
+    await decorateTemplate(main);
```
mfrisbey, line 202: For my benefit, what is the value of moving this method call here instead?

mfrisbey (approved):
Approved - I did leave some questions, but nothing that would block a merge.

PR #78, opened 2023-05-11
flesh out pc resolution for article template

### LINKED ISSUE #8: feat: implement template system for the various pages

mfrisbey:

Fix #8 

Test URLs:
- After: https://issue-8-article-full--petplace--hlxsites.hlx.page/

This PR removes a couple of blocks, in favor of building the elements inline (this cleans up the DOM). Adds styling for the largest resolution layout of the article template.


ramboz (approved):
Looking good! 🚀

PR #79, opened 2023-05-12
Issue 38-  Slide Cards

### LINKED ISSUE #38: feat: breed facts slide block
Image
Image

### LINKED ISSUE #39: feat: other breeds slider block variant
Image
Image

travrob1:
cards used for both facts and different breeds cards

Fix #38 
Fix #39

Test URLs:
- After: https://issue-38--petplace--hlxsites.hlx.page/article/breed/pembroke-welsh-corgi

image

image


blocks/slide-cards/slide-cards.css
```diff
@@ -0,0 +1,174 @@
...
+
+.slide-cards {
+    list-style: none;
+    display: flex;
+    padding: 0;
+}
+
+.slide-card {
+    background-color: var(--background-color);
+    color: #4a4a4a;
```
ramboz, line 25:
```suggestion
    color: var(--color-gray-dark);
```

blocks/slide-cards/slide-cards.css
```diff
@@ -0,0 +1,174 @@
...
+    font-size: 1em;
+    font-weight: 400;
+    line-height: 1.5;
+    user-select: none;
+    -webkit-tap-highlight-color: transparent;
+    min-height: 300px;
+    margin: 10px 45px 100px;
```
ramboz, line 32: *nitpicking*: you could extract the `45px` as a variable and thus re-use it in the width calculation so the links between the 2 is more explicit

blocks/slide-cards/slide-cards.css
```diff
@@ -0,0 +1,174 @@
...
+
+.numbered .slide-card {
+    padding: 0 30px 30px;
+}
+
+.numbered .slide-card h4 {
+    padding-top: 36px;
+    font-size: 80px;
+    position: relative;
+    font-family: Raleway,sans-serif;
```
ramboz, line 52:
```suggestion
    font-family: var(--heading-font-family);
```

blocks/slide-cards/slide-cards.js
```diff
@@ -0,0 +1,43 @@
+import { slide, initializeTouch } from '../../templates/breed-page/breed-page.js';
+
+export default async function decorate(block) {
+  const breedColorClasses = ['breed-purple', 'breed-red', 'breed-blue'];
+
+  [...block.children].forEach((child, i) => {
+    if (i === 0) child.setAttribute('active', true);
+
+    if (block.classList.contains('numbered')) {
```
ramboz, line 9: Could we do this in pure CSS? Leveraging the [CSS counters](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Counter_Styles/Using_CSS_counters)

templates/breed-page/breed-page.css
```diff
@@ -43,6 +45,57 @@ main .section.centered p {
...
+  outline: none;
+  background: var(--color-gray-light-alt);
+  height: 48px;
+  width: 48px;
+  border-radius: 24px;
+  display: block;
+}
+
+.slick-prev {
+  left: calc(50% - 80px);
```
ramboz, line 64: That `80px` feels like a magic number. Is this stable enough when you resize? and is that value derived from a padding/margin maybe?

templates/breed-page/breed-page.css
```diff
@@ -43,6 +45,57 @@ main .section.centered p {
...
+  vertical-align: middle;
+  transform: rotate(180deg);
+}
+
+.slick-next {
+  right: calc(50% - 80px);
+  z-index: 99;
+}
+
+.slick-next::before {
```
ramboz, line 86: Lots of properties in common with `.slick-prev::before`, maybe we can combine the 2 and only specify the overrides here

templates/breed-page/breed-page.js
```diff
@@ -19,3 +19,59 @@ export async function buildTemplateBlock(main) {
   await decorateIcons(p);
   h1.insertAdjacentElement('afterend', p);
 }
+
+export function getActiveSlide(block) {
```
ramboz, line 23: Doesn't look like you are actually using it anywhere else, so we can maybe remove the export

templates/breed-page/breed-page.js
```diff
@@ -19,3 +19,59 @@ export async function buildTemplateBlock(main) {
...
+  return {
+    index: [...block.children].findIndex((child) => child.getAttribute('active') === 'true'),
+    element: block.querySelector('[active="true"]'),
+    totalSlides: [...block.children].length,
+  };
+}
+
+export function slide(slideDirection, block, slideWrapper) {
```
ramboz, line 31: I'd be tempted to move this to a "virtual" block instead, or a dedicated JS file. It's not necessarily limited to the breeds pages and could theoretically be used everywhere

blocks/slide-cards/slide-cards.js
travrob1: I'm really confused what you are asking for.  the `numbered` is just the subtype of slide card
image

We need javascript logic to inject the necessary html into these different subtypes of slide cards.

templates/breed-page/breed-page.css
travrob1: the 80px offset is what adds spacing in between the next and previous buttons.. it does not need to adjust responsivly.

image

templates/breed-page/breed-page.css
travrob1: good call!

templates/breed-page/breed-page.js
travrob1: a dedicated js file? i thought i would just put it into the scripts.js file if not in the tempalte... Where else should i put it?

blocks/slide-cards/slide-cards.js
ramboz: You need JS to decorate, yes, but injecting the numbers themselves could be done via CSS. So we save ourselves a few lines of JS

templates/breed-page/breed-page.css
ramboz: ok, if it's just to have a stable 160px spacing between the 2 we are all good :) I was wondering if this was proportional to another value and could be `calc`'ed but it doesn't seem so.

templates/breed-page/breed-page.js
ramboz: `scripts.js` could also work. it just feels like you started a "meta block" or "virtual block" that centralizes some logic around sliding, so thought we could create a dedicated JS for this. Can also later be easily re-used in other projects that way

blocks/slide-cards/slide-cards.js
travrob1: seems to not be a big deal... lest just leave it and i will note the css counters for future use ;)

blocks/slide-cards/slide-cards.css
```diff
@@ -0,0 +1,174 @@
...
+    list-style: none;
+    display: flex;
+    padding: 0;
+}
+
+.slide-card {
+    background-color: var(--background-color);
+    color: var(--color-gray-dark);
+    font-size: 1em;
+    font-weight: 400;
```
ramboz, line 27: Do we need the font-weight and line-height? aren't the inherited props enough?

ramboz (approved):

blocks/slide-cards/slide-cards.css
travrob1: nope.. good call.

PR #83, opened 2023-05-16
Issue 16

### LINKED ISSUE #16: feat: categories and color mappings

### LINKED ISSUE #81: feat: article bread crumbs
Dynamically add article breadcrumbs.

Image

travrob1:
Colors for:
Home page tiles
Category Pages
Breadcrumbs

Fix #16 
Fix #81 

Test URLs:
- After: https://issue-16--petplace--hlxsites.hlx.page/

Screen Shot 2023-05-16 at 3 59 36 PM
Screen Shot 2023-05-16 at 3 59 47 PM
Screen Shot 2023-05-16 at 3 59 58 PM
Screen Shot 2023-05-16 at 4 00 23 PM


scripts/scripts.js
```diff
@@ -297,10 +298,13 @@ async function loadLazy(doc) {
  * Loads everything that happens a lot later,
  * without impacting the user experience.
  */
-function loadDelayed() {
+function loadDelayed(doc) {
   // eslint-disable-next-line import/no-cycle
   window.setTimeout(() => import('./delayed.js'), 3000);
   // load anything that can be postponed to the latest here
+  if (templateModule?.loadDelayed) {
```
ramboz, line 305: I would move this into the `setTimeout`. The delayed is supposed to execute after 3s, this guarantees that the page has stabilized and we don't affect the rendering of the critical stuff

templates/article-page/article-page.js
```diff
@@ -75,3 +108,41 @@ export function buildTemplateBlock(main) {
...
+      const chevron = document.createElement('span');
+      chevron.innerHTML = '<span class="icon icon-chevron"></span>';
+      breadCrumbs.append(chevron);
+    }
+    const linkButton = document.createElement('a');
+    linkButton.href = crumb.url;
+    linkButton.innerText = convertToTitleCase(crumb.path);
+    linkButton.classList.add('category-link-btn');
+    if (i === crumbData.length - 1) {
+      linkButton.classList.add(`${color}`);
```
ramboz, line 139: I'd be tempted to try something like:
```css
.category-link-btn {
  --bg-color: inherit;
  --border-color: inherit;
  --text-color: inherit;
  background: var(--bg-color);
  border: 1px solid var(--border-color);
  color:  var(--text-color);
}
```
```js
if (i === crumbData.length - 1) {
    linkButton.style.setProperty('--bg-color', `var(--color-${color})`);
    linkButton.style.setProperty('--border-color', 'inherit');
    linkButton.style.setProperty('--text-color', 'inherit');
} else {
    linkButton.style.setProperty('--bg-color', 'inherit');
    linkButton.style.setProperty('--border-color', `var(--color-${color})`);
    linkButton.style.setProperty('--text-color', `var(--color-${color})`);
}
```

The idea is that we'd set local variables for the stuff that we want to dynamically change, and then leverage the JS API to inject the right value at runtime based on the category we have.

If it works, you can essentially then do away with all the custom CSS classes for each color and we just stick to the variables

travrob1: This works if there is one element on the page that needs to use --bg-color... but what if you have multiple elements?  Then you will run into conflicts and be back to the class names.

ramboz: I don't think so, because you override the properties on the element itself, and not globally on all styles.

travrob1: you are indeed correct sir.  adjusting the rest of the site

ramboz: 🙌🏼 This will simplify things quite a bit!
Wasnt sure it would work, but this could come in handy in other projects 😄

scripts/scripts.js
travrob1: done

ramboz (approved):

PR #92, opened 2023-05-17
Implement article navigation block

### LINKED ISSUE #44: feat: previous/next article navigation links
Image

mfrisbey:
Fix #44 

Test URLs:
- After: https://issue-44--petplace--hlxsites.hlx.page/

Here is an example page with the navigation: https://issue-44--petplace--hlxsites.hlx.page/article/cats/pet-care/best-halloween-themed-cat-names

Here is an example page _without_ the navigation: https://issue-44--petplace--hlxsites.hlx.page/article/dogs/pet-care/dog-care/grooming-your-dog/the-ultimate-how-to-guide-for-grooming-a-dog-at-home

The algorithm for displaying next previous is:

* Get `/categories.json` and filter for all articles in the current article's category.
* Sort the articles by the date they were created.
* Find the current article in the category list.
* The previous article will be the next older article in the list (it will wrap around to the end if the current article is the first one)
* The next article will be the next newer article in the list (it will wrap around to the beginning if the current article is the last one)

The article navigation won't be displayed at all if there are less than 3 articles in the category (i.e. there aren't enough for the previous, current, and next).

In terms of layout, it uses a CSS grid to position all the items in the block. As the window width grows, the space between the previous/next card will also grow, up to the section maximum width.

The block changes quite a bit on mobile devices, so I'd recommend switching device types to see the need for some of the extra divs.


blocks/article-navigation/article-navigation.css
```diff
@@ -1 +1,182 @@
...
+}
+
+.article-navigation svg {
+  width: 19px;
+  height: 47px;
+}
+
+.article-navigation-combined-header, .article-navigation-next-header, .article-navigation-previous-header {
+  font-size: 1.125rem;
+  font-weight: 800;
```
ramboz, line 21:
```suggestion
  font-size: var(--heading-font-size-m);
```

blocks/article-navigation/article-navigation.css
```diff
@@ -1 +1,182 @@
...
+  border-radius: 6px;
+}
+
+.article-navigation-category a {
+  color: var(--text-color);
+}
+
+.article-navigation-category {
+  font-weight: 800;
+  font-size: .75rem;
```
ramboz, line 69: I'd be tempted to create a new `--heading-font-size-xxs: .75em;` in styles.css and use that instead

blocks/article-navigation/article-navigation.css
```diff
@@ -1 +1,182 @@
...
+  align-self: center;
+}
+
+@media (min-width: 1024px) {
+  .article-navigation-combined-header, .article-navigation-next-header, .article-navigation-previous-header {
+    font-size: 1.375rem;
+  }
+
+  .article-navigation-category {
+    font-size: .875rem;
```
ramboz, line 115:
```suggestion
    font-size: var(--heading-font-size-xs);
```

blocks/article-navigation/article-navigation.css
```diff
@@ -1 +1,182 @@
...
+    line-height: 1.5;
+    font-weight: 400;
+    font-family: unset;
+  }
+
+  .article-navigation-title {
+    font-size: 1.75rem;
```
ramboz, line 122:
```suggestion
    font-size: var(--heading-font-size-xl);
```

blocks/article-navigation/article-navigation.css
```diff
@@ -1 +1,182 @@
...
+  .article-navigation-previous-title-text, .article-navigation-next-title-text {
+    max-width: 280px;
+  }
+
+  .article-navigation {
+    gap: 1.5rem;
+    grid-template:
+      ". previous previous-header previous-details next-header next-details next ." auto
+      ". previous previous-thumbnail previous-details next-thumbnail next-details next ." 1fr
+      / auto auto auto 1fr auto auto auto auto;
```
ramboz, line 144: I'd be tempted to go with an additional "spacer" in the middle:
```suggestion
      ". previous previous-header previous-details .. next-header next-details next ." auto
      ". previous previous-thumbnail previous-details .. next-thumbnail next-details next ." 1fr
      / auto min-content min-content min-content 1fr min-content min-content min-content auto;
```

blocks/article-navigation/article-navigation.js
```diff
@@ -1,3 +1,155 @@
...
+  if (!articles) {
+    const res = await fetch('/article/query-index.json?sheet=article');
+    const queryData = await res.json();
+    articles = queryData?.data;
+  }
+
+  // find articles in the same category, sorting by creation date ascending
+  const similarArticles = articles
+    .filter((article) => article.category === category)
+    .sort((a, b) => {
```
ramboz, line 69: We could maybe do this server side with a new sheet that is properly sorted by publication date, or just update the existing one so it is. On >2K articles, I guess this would lead to some TBT otherwise.

ramboz (approved):
Looks good overall, just a few minor comments

blocks/article-navigation/article-navigation.css
mfrisbey: `--heading-font-size-xs` value of `.875em` is _slightly_ different than `.875rem` and doesn't exactly match their current site. Is that ok?

blocks/article-navigation/article-navigation.css
mfrisbey: Same here, but the difference between `1.75rem` and `1.75em` is more obvious.

blocks/article-navigation/article-navigation.css
mfrisbey: This is what I originally had. It's cleaner and I felt like it worked better, but it added extra spacing between the columns due to the gap of the spacer column. I'll add the spacer back and we'll see what feedback they have.

mfrisbey: P.S. retaining `auto` worked better than `min-content` on the other columns.

blocks/article-navigation/article-navigation.js
```diff
@@ -1,3 +1,155 @@
...
+  sectionContainer.classList.add('article-navigation-details', `article-navigation-${key}-details`);
+  sectionContainer.append(category);
+  sectionContainer.append(title);
+
+  block.append(thumbnail);
+  block.append(sectionContainer);
+}
+
+async function createNavigation(block) {
+  const category = getMetadata('category');
```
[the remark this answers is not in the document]
mfrisbey: I originally had it this way, but noticed the category in the metadata was different than the category in the URL. Which should be higher priority? I'm assuming it would be:

* Use the category from metadata first.
* If no metadata, use category from URL.

Is that correct?

blocks/article-navigation/article-navigation.js
mfrisbey: I like the idea of pulling it in whichever order they are in the existing article sheet. That way then can control/change the order if they want without needing to update the code. I went ahead and sorted the existing `raw_index` tab in the `query-index.xlsx` file by date in ascending order, and I removed the sort from the code.

Is this something we should document somewhere?

blocks/article-navigation/article-navigation.js
```diff
@@ -1,3 +1,155 @@
...
+  const nextHeader = document.createElement('div');
+  nextHeader.classList.add('article-navigation-next-header');
+  nextHeader.innerText = 'Next Article';
+  block.append(nextHeader);
+
+  const nextArticle = similarArticles[nextIndex];
+  const previousArticle = similarArticles[prevIndex];
+
+  // left arrow
+  const leftNav = document.createElement('div');
```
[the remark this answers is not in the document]
mfrisbey: Do you have an example? I haven't used `::before` very much, and would like to learn.

mfrisbey:
@ramboz I changed most everything, except the font sizes (because the units are different), and your `::before` suggestion (because I wasn't sure how to do that). I left several questions, please feel free to take another look and let me know your thoughts.

blocks/article-navigation/article-navigation.css
ramboz: hmm… ideally they should match. we may have some inheritance issue somewhere. just keep what you had then, we can always refactor later if we really want to

blocks/article-navigation/article-navigation.css
ramboz: Same as above

blocks/article-navigation/article-navigation.js
ramboz: I think we'll need to sort it using the SORT+FILTER functions for it to stay updated with new entries. Not sure… we'll have to test it out a bit.

We can document this at the end when we do the handover and training. Things might change again until then

blocks/article-navigation/article-navigation.js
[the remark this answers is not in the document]
ramboz: Yes, exactly!

blocks/article-navigation/article-navigation.js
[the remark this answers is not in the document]
ramboz: You could probably do something like:
```html
<div class="prev-wrapper>
  …
</div>
<div class="next-wrapper>
  …
</div>
```

```css
.prev-wrapper::before,
```

PR #93, opened 2023-05-18
feat: add import logic for article pages

### LINKED ISSUE #28: chore: import content from sitemap
Ask on [#franklin-import-as-a-service](https://cq-dev.slack.com/archives/C04RN9Z4HSP) for import of https://www.petplace.com/sitemap-0.xml

Sub-tasks:
- [x] replace all "https://petplace.uat.petpartners.com/" links in the content with relative [petplace.com](http://petplace.com/) links

ramboz:

Fix #28 

Test URLs:
- After: https://issue-28--petplace--hlxsites.hlx.page/


tools/importer/import.js
```diff
@@ -0,0 +1,331 @@
...
+
+  return meta;
+};
+
+function getTemplate(url) {
+  const { pathname } = new URL(url);
+  if (pathname === '/') {
+    return 'home-page';
+  }
+  if (pathname.startsWith('/article/breed/')) {
```
travrob1, line 49: the breed index uses a different template then the breed page.  this logic will return true on the breed index. If the breed index page is excluded from the importer than this is a none issue.

ramboz: Yes, this is still WIP and I've only looked at the article and breed pages for now, but didn't look at the index pages yet.

PR #96, opened 2023-05-19
Add author template

### LINKED ISSUE #84: feat: add the authors page
Some authors actually have a dedicated page. For instance: https://www.petplace.com/authors/dr-debra-primovic-dvm

We need to add those:

- [x] Manually import the page into sharepoint
- [x] Create a dedicated template for the page
- [x] Re-use the social links block from the article page
- [ ] Re-use the "other breeds" block from breeds pages for the popular posts

mfrisbey:
Fix #84 

Test URLs:
- After: https://issue-84--petplace--hlxsites.hlx.page/

This PR adds a template for the author bio pages. In summary, this PR:

* Adds a new `author-page` template and styling for all resolutions.
* Moves the breadcrumb styling and javascript into shared locations. The author pages also have breadcrumbs, but all the breadcrumb functionality was in the article template. The shared location allows usage of breadcrumbs anywhere in the site.

Here is an example author page: https://issue-84--petplace--hlxsites.hlx.page/authors/carey-hemmelgarn-dvm-dacvecc


scripts/lib-franklin.js
```diff
@@ -606,6 +606,71 @@ export function setup() {
   }
 }
 
+/**
+ * Convert snake case to title case
+ * @param str
+ * @returns {*}
+ */
+function convertToTitleCase(str) {
```
ramboz, line 614: I wouldn't change `lib-franklin.js` too much as it will make it harder to update later against the boilerplate.
Custom project code is better stored in `scripts.js`

scripts/lib-franklin.js
```diff
@@ -606,6 +606,71 @@ export function setup() {
...
+
+/**
+ * Adds the given breadcrumb data to an HTML element, styling it for
+ * use in the site.
+ * @param {Element} target HTML element to which the breadcrumbs will be
+ *  added.
+ * @param {Array<CrumbData>} crumbData Information about the crumbs to add.
+ * @returns {Promise} Resolves when the crumbs have been added.
+ */
+export async function buildBreadCrumbs(target, crumbData) {
```
ramboz, line 636: I would propose extracting this to a breadcrumbs block instead and then leverage the `buildBlock` and `loadBlock` APIs from `lib-franklin.js` to trigger the decoration.

templates/author-page/author-page.js
```diff
@@ -0,0 +1,39 @@
...
+  main.append(breadcrumb);
+  createTemplateBlock(main, 'social-links');
+  createTemplateBlock(main, 'popular-articles');
+}
+
+export async function loadLazy(main) {
+  const heading = main.querySelector('h1');
+  const breadCrumbs = main.querySelector('.author-template-breadcrumb');
+
+  buildBreadCrumbs(breadCrumbs, [{
```
ramboz, line 30: If you use a block for breadcrumbs, this would essentially turn into something like:
```js
const ol = document.createElement('ol'); // important for accessibility, the list needs to be ordered
[ /* home */, /* authors */, /* author */].forEach((item) => {
  const li = document.createElement('li');
  …
  ol.append(li);
});
const crumbs = buildBlock('hero', { elems: [ol] }))
loadBlock(crumbs);
```

scripts/lib-franklin.js
mfrisbey: Thanks for the tip, I didn't realize that. I'll move things around.

scripts/lib-franklin.js
mfrisbey: I had originally wanted to make a block out of it, but there is this code in scripts.js for the article pages that adds the breadcrumbs to the hero block:

```
      const breadcrumb = document.createElement('div');
      breadcrumb.classList.add('breadcrumb', 'article-template-breadcrumb');
      section.append(buildBlock('hero', { elems: [optimized, h1, breadcrumb] }));
```

Would be be safe to use `buildBlock('breadcrumb')` here and still pass it as a child element of another block? I remember hearing that having blocks within blocks isn't recommended, so I'm hoping to understand the rules there.

ramboz: I think the concern about blocks inside of blocks is mostly for the authoring side since you end up with embedded tables that are hard to maintain. From the code side, it's less of an issue, especially for "virtual blocks" that aren't actually authored, but just auto-generated.

We could probably cleanup the `scripts.js` and move the logic inside the template logic. Especially now that they support lazy/eager phases. The template could inject the breadcrumb in the lazy into the header that was already rendered. Especially with a revered flex column layout, the new breadcrumb element would just naturally stack on top of the heading async and avoid any CLS.

Does that make sense?

mfrisbey:
@ramboz I took another pass at this based on your feedback. I created a new block for the breadcrumbs, and added a shared backend method that builds the element that go into the block. The articles template uses the method when lazy loading to create the categories (since it's more expensive and requires reading the category data); the author template uses the method during the eager phase, because it's a static category structure.

I noticed that the existing breadcrumbs don't use `ol` as you suggested. I didn't change that as part of this PR because I didn't want to add too much to it. Is that something that will need to be modified eventually?

ramboz:
@mfrisbey Yes, we can tackle this in #102

ramboz:
Adding the breadcrumb currently adds quite a bit of CLS:
Screenshot 2023-05-22 at 5 36 50 PMScreenshot 2023-05-22 at 5 36 56 PM

You can fix this by pre-setting the height/margin on the hero and then placing the breadcrumb via position absolute later.

PR #101, opened 2023-05-19
Issue 33 - Breed Cards and Filters

### LINKED ISSUE #33: feat: breed filter
Image

travrob1:
Breed Cards, filters, and pagintation

Screen Shot 2023-05-19 at 3 23 04 PM

Screen Shot 2023-05-19 at 3 23 47 PM

Fix #33 

Test URLs:
- Before: https://main--petplace--hlxsites.hlx.page/article/breed/
- After: https://issue-33--petplace--hlxsites.hlx.page/article/breed/

Built the page and logic with hard coded test data.  Now calling the query-index and getting back a single breed... for now


blocks/type-filters/type-filters.js
```diff
@@ -0,0 +1,152 @@
...
+
+  block.innerHTML = `
+      <div class="type-filter">
+        <div class="filter-btn"><span>Filters</span></div>
+        <div class="category-filters">
+          <h3>
+            Type <span class="icon icon-chevron"></span>
+          </h3>
+          <div class="filter-type is-active">
+            <div class="checkbox-wrapper"><input type="checkbox" checked="" name="Sporting"><span><label for="Sporting">Sporting</label></span></div>
```
ramboz, line 88: You could simplify this slightly by creating an array of types (`['Sporting', 'Hound', …]` and then doing a `.map(…).join('')` in here to get them all generated at once

ramboz (commented):
I'm thinking we should decouple the filters, the paginations and the list in separate blocks each. Each would only control the browser URL by injecting new query parameters or updating them, and the list would listen to those changes and automatically update.

- Filters => essentially set the `&type=` query parameter
- Pagination => essentially adds the `&=offset=` or `&page=` parameter and that drives the `offset` param on the query to the json

We can do those by using the `history.pushState` API.

This will open up back/forward navigation which makes the site more accessible (as you can easily rewind to previous states), and it would also make sharing Urls easier

blocks/type-filters/type-filters.css
```diff
@@ -0,0 +1,239 @@
...
+    color: var(--text-color);
+    box-sizing: inherit;
+    overflow: hidden;
+    transition: all .35s;
+    display: block;
+    max-height: 0;
+}
+
+.category-filters.is-active {
+    max-height: 650px;
```
ramboz, line 42: What drives this exact value? I guess if devs wanted to add more breed types down the road, we'd need to adjust this accordingly. Maybe add a quick comment with the calculation you did

blocks/type-filters/type-filters.css
```diff
@@ -0,0 +1,239 @@
...
+}
+
+.category-filters h3 {
+    display: block;
+    width: calc(100% - 1rem);
```
ramboz, line 47: A heading should already take the whole width by default

blocks/type-filters/type-filters.css
```diff
@@ -0,0 +1,239 @@
...
+    padding: 0;
+}
+
+.filter-type {
+    padding: 0 .5rem;
+    transition: all .5s;
+    max-height: 0;
+}
+
+.filter-type.is-active {
```
ramboz, line 101: This filters seems to act as a single accordion. You might want to reuse the `aria-accordion.js` to decorate this automatically and essentially get an accessible widget. This will also save you the JS toggle logic to some extent and a few CSS selectors

blocks/type-filters/type-filters.css
```diff
@@ -0,0 +1,239 @@
...
+    max-height: 400px;
+}
+
+.card-container {
```
ramboz, line 105: The grid and container will likely conflict with @varunmitra's #104

blocks/type-filters/type-filters.css
```diff
@@ -0,0 +1,239 @@
...
+    border-top-left-radius: 8px;
+    border-top-right-radius: 8px;
+}
+
+.grid-item h3{
+    text-align: left;
+    margin: 20px 20px 40px;
+}
+
+.pagination {
```
ramboz, line 163: Same for the pagination

blocks/type-filters/type-filters.js
```diff
@@ -0,0 +1,165 @@
...
+  });
+
+  // Add click listener for the category  filter chevron button
+  const chevron = block.querySelector('.icon-chevron');
+  chevron.addEventListener('click', () => {
+    const filterType = block.querySelector('.filter-type');
+
+    if (filterType.classList.contains('is-active')) {
+      filterType.classList.remove('is-active');
+      chevron.style.setProperty('transform', 'rotateX(0deg)');
```
ramboz, line 136: You can probably do those directly in CSS since they don't need a dynamic value.

PR #104, opened 2023-05-22
Issue 29

### LINKED ISSUE #29: feat: cards block
Image
Image

varunmitra:

Fix #29 

Test URLs:
- Before: https://main--petplace--hlxsites.hlx.page/article/category/pet-care/cat-care/ 
- After: https://issue-29--petplace--hlxsites.hlx.page/article/category/pet-care/cat-care/


blocks/blog-cards/blog-cards.css
```diff
@@ -0,0 +1,174 @@
...
+    display: block;
+   
+    
+  }
+   
+  .blog-cards > ul > li img {
+    border-top-left-radius: 6px;
+    border-top-right-radius: 6px;
+    aspect-ratio: 4 / 3;
+    max-width: 100%;
```
ramboz, line 46: you'll probably want to add `object-fit: cover` so the images are not distorted

blocks/blog-cards/blog-cards.js
```diff
@@ -0,0 +1,119 @@
...
+
+function buildPost(post, eager, childCategories) {
+  let category;
+  let categorypath;
+  let categorycolor;
+  const postCard = document.createElement('div');
+  postCard.classList.add('blog-cards');
+  const postDate = new Date(0);
+  postDate.setUTCSeconds(post.date);
+  const postDateStr = postDate.toLocaleString('default', { month: 'long' }).concat(' ', postDate.getDate(), ', ', postDate.getFullYear());
```
ramboz, line 25: The TBT we have is likely due to this. The Locale APIs are notoriously really slow.
I'd suggest decorating the page first with a placeholder value and then coming back async with the final formatted value.

You can take a look at https://github.com/hlxsites/petplace/blob/main/blocks/article-author/article-author.js#L11-L15

blocks/blog-cards/blog-cards.js
```diff
@@ -0,0 +1,119 @@
...
+  postCard.innerHTML = `
+      <div class="blogs-card-image">
+        <a href="${post.path}">${createOptimizedPicture(post.image, `Teaser image for ${post.title}`, eager).outerHTML}</a>
+        <a aria-current="page" class="blogs-card-category ${categorycolor}" href="${categorypath}" data-acsb-clickable="true" data-acsb-navigable="true" data-acsb-now-navigable="true">${category}</a>
+      </div>
+      <div>              
+        <a href="${post.path}">
+        <div class="blogs-card-body">
+        <h3>${post.title}</h3>
+        <p><span class="card-date">${postDateStr.concat(' · ', post.author)}</span></p>
```
ramboz, line 42: I would leverage the `<time datetime="…">…</time>` instead to have better semantics out of the box

blocks/blog-cards/blog-cards.css
```diff
@@ -0,0 +1,174 @@
...
+    font-size: 1em;
+    font-weight: 400;
+    line-height: 1.5;
+    display: block;
+    padding-inline-start: 0;
+    min-height: 100vh;
+  }
+  
+  .blog-cards > ul > li {     
+    border: 1px solid #c4d1e9;
```
ramboz, line 18: We also encountered this color in `article-author.css`, so might be a good opportunity to extract it as a variable into `styles.css`, maybe as `--color-blue-light`

blocks/blog-cards/blog-cards.css
```diff
@@ -0,0 +1,174 @@
...
+    border-radius: 6px;
+    padding: 0;
+    margin-bottom: 1.5rem;    
+  }
+  
+  .blog-cards .blogs-card-body {
+    padding: 24px;
```
ramboz, line 25: This could be set to `1.5rem` so it grows proportionally

blocks/blog-cards/blog-cards.css
```diff
@@ -0,0 +1,174 @@
...
+    display: flex;    
+    flex-direction: column;
+    -ms-flex-pack: justify;
```
ramboz, line 28: I'd avoid the ms specific prefixes and stick to the official flex properties

blocks/blog-cards/blog-cards.css
```diff
@@ -0,0 +1,174 @@
...
+@media (min-width: 400px) {
+.blog-cards > ul {
+    list-style: none;
+    margin: 0;                    
+    font-size: 1em;
+    font-weight: 400;
+    line-height: 1.5;
+    display: block;
+    padding-inline-start: 0;
+    min-height: 100vh;
```
ramboz, line 14: I'm not sure this is needed and would extend too high if we only have a few cards (thinking about the last pagination pages)

blocks/blog-cards/blog-cards.css
```diff
@@ -0,0 +1,174 @@
+main a {
+    color: #4a4a4a;
```
ramboz, line 2:
```suggestion
    color: var(--color-gray-dark); 
```

blocks/blog-cards/blog-cards.css
ramboz, line 1: I'd be tempted to just name these `Cards` and essentially re-use the existing cards block from the boilerplate. I don't think we are using it anywhere else at the moment

blocks/blog-cards/blog-cards.css
```diff
@@ -0,0 +1,174 @@
...
+  }
+   
+  .blog-cards > ul > li img {
+    border-top-left-radius: 6px;
+    border-top-right-radius: 6px;
+    aspect-ratio: 4 / 3;
+    max-width: 100%;
+  }
+  
+  .blog-cards .blogs-card-body .h3{
```
ramboz, line 49: This wouldn't be used, we don't have explicit instrumentation of the heading classes. At best it would be plain `h3` but I guess most of the styles would anyway be inherited

blocks/blog-cards/blog-cards.css
```diff
@@ -0,0 +1,174 @@
...
+    font-weight: 800;
+    color: #28061e;
+  }
+
+  .blog-cards .blogs-card-body >span{
+    font-size: var(--body-font-size-s);
+  }
+
+  .blog-cards .blogs-card-category{
+    font-family: Raleway,sans-serif;
```
ramboz, line 62:
```suggestion
    font-family: var(--heading-font-family);
```

blocks/blog-cards/blog-cards.css
```diff
@@ -0,0 +1,174 @@
...
+    font-weight: 800;
```
ramboz, line 63: not needed as it is inherited from the font face declaration

blocks/blog-cards/blog-cards.css
```diff
@@ -0,0 +1,174 @@
...
+    position: absolute;
+    line-height: 1.7;
+    top: 10px;
+    left: 20px;
+    border-radius: 14px;
+    display: inline-block;
+    padding: 0 14px;    
+    color: #fff;
```
ramboz, line 71:
```suggestion
    color: var(--text-color-inverted);
```

blocks/blog-cards/blog-cards.css
```diff
@@ -0,0 +1,174 @@
...
+    max-width: calc(100% - 2rem);
```
ramboz, line 72: In theory, I'd set an `offset` variable to your left position (`20px`) and calc based on twice that variable so the relationship is clear. As it is, if 1rem ended up being larger than 20px then it wouldn't render properly and overlap slightly… really more of a theoretical edge case than a real problem, but using variables makes the relationship explicit and easier to maintain long-term

blocks/blog-cards/blog-cards.css
```diff
@@ -0,0 +1,174 @@
...
+  }
+
+  .blog-cards .pagination{
+    margin-top: 2rem;
+    font-size: 1.125rem;
+    line-height: 32px;      
+    justify-content: flex-end;
+    padding: 0 0.75rem;    
+    display: flex;
+    -ms-flex-preferred-size: 100%;
```
ramboz, line 82: Let's remove the ms prefixed props :)

blocks/blog-cards/blog-cards.css
```diff
@@ -0,0 +1,174 @@
...
+}
+
+.blog-cards .pagination >li .prev::before {
+  position: absolute;
+  left: 0;
+  top: 0;
+  content: "";    
+  width: 100%;
+  height: 100%;
+  background: url('../../icons/previous.svg');
```
ramboz, line 109: I'd use a neutral colored svg icon and color it locally using background + mask, like we did here for instance:
https://github.com/hlxsites/petplace/blob/6a846cb063d6abd0a93b44bdbb7397298dc5cbad/styles/lazy-styles.css#L137-L140

blocks/blog-cards/blog-cards.css
```diff
@@ -0,0 +1,174 @@
...
+}
+
+  .blog-cards .pagination >li .next{
+    position: relative;
+    width: 7px;
+    height: 32px;
+    margin-left: 1rem;
+  }
+
+  .blog-cards .pagination >li .next::after {
```
ramboz, line 122: Most of the styles are the same as on the `.prev::before` selector, so we could probably combine those and just override what's needed specifically for the second

blocks/blog-cards/blog-cards.css
```diff
@@ -0,0 +1,174 @@
...
+@media (min-width: 900px) {
+   .blog-cards > ul {      
+      display: grid;
+      grid-template-columns: repeat(3, 31.33%);
+      grid-gap: 24px;   
+      margin-bottom: 1rem;             
+    }
+
+    
+.acsb-sr-only {
```
ramboz, line 145: You can skip those, these will be added by the accessibility widget itself if needed.

blocks/blog-cards/blog-cards.css
```diff
@@ -0,0 +1,174 @@
...
+  background-size: contain;
+  background-position: 50%;
+  background-repeat: no-repeat;
+}
+}
+
+@media (min-width: 900px) {
+   .blog-cards > ul {      
+      display: grid;
+      grid-template-columns: repeat(3, 31.33%);
```
ramboz, line 139: You could probably just use `grid-template-columns: repeat(3, 1fr);` so columns are sized at the same size

blocks/blog-cards/blog-cards.css
```diff
@@ -0,0 +1,174 @@
...
+  border: 0 !important;
+  outline: 0 !important;
+  cursor: text !important;
+  text-transform: capitalize !important;
+  color: #000 !important;
+  background: #fff !important;
+}
+  }
+
+  @media (min-width: 1200px) {
```
ramboz, line 168: I'd skip those and keep what you had on the 900px breakpoint. the grid columns would be the same, and I don't see the UX improvement of reducing spacing on larger viewports. Intuitively I'd even go and increase those instead.

blocks/blog-cards/blog-cards.js
```diff
@@ -0,0 +1,119 @@
+import ffetch from '../../scripts/ffetch.js';
+import { createOptimizedPicture, decorateIcons } from '../../scripts/lib-franklin.js';
+
+async function fetchCategoryData() {
```
ramboz, line 4: You could re-use the `getCategories`/`getCategory` we defined in `scripts.js`. There aren't enough to require pagination, and once they are fetched we also cache them locally for future calls in the session storage.

blocks/blog-cards/blog-cards.js
```diff
@@ -0,0 +1,119 @@
...
+    if (item.Category === post.category) {
+      category = item.Category;
+      categorypath = item.Path;
+      categorycolor = item.Color;
+    }
+  });
+  postCard.innerHTML = `
+      <div class="blogs-card-image">
+        <a href="${post.path}">${createOptimizedPicture(post.image, `Teaser image for ${post.title}`, eager).outerHTML}</a>
+        <a aria-current="page" class="blogs-card-category ${categorycolor}" href="${categorypath}" data-acsb-clickable="true" data-acsb-navigable="true" data-acsb-now-navigable="true">${category}</a>
```
ramboz, line 36: All the `data-acsb-*` attributes aren't really needed. those are injected by their accessibility widget when enabled.

You can also remove `aria-current="page"`, this wasn't used properly on their original site. I don't think we are in any of the use cases described here: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-current

blocks/blog-cards/blog-cards.js
```diff
@@ -0,0 +1,119 @@
...
+    }
+    const li = document.createElement('li');
+    li.append(buildPost(post, i < 1, categoriesData));
+    newUl.append(li);
+    i += 1;
+  }
+  pagesElem.innerHTML = `
+      <ul class="pagination">
+        <li class="prev"><a data-page="${pageNum - 1}" href="${window.location.pathname}?page=${pageNum}"><span class="icon icon-previous"><span class="sr-only">Previous Page</span></a></li>
+        <li><span>Page ${pageNum + 1}</span></li>
```
ramboz, line 77: The span here could get the `aria-current="page"` attribute on the other hand since it does represent the current active page in the pagination.

blocks/blog-cards/blog-cards.js
```diff
@@ -0,0 +1,119 @@
...
+  if (pageNum === 0) {
+    pagesElem.querySelector('.prev').remove();
+  }
+  if (!morePages) {
+    pagesElem.querySelector('.next').remove();
+  }
+  pagesElem.querySelectorAll('li > a').forEach((link) => {
+    link.addEventListener('click', (evt) => {
+      evt.preventDefault();
+      buildBlogFeed(ul, Number(link.dataset.page), pagesElem);
```
ramboz, line 90: I would also push the offset or page index to the URL as a query parameter via the `history.pushState(…)` API and base the initial index on that query parameter if present. That way the individual pages can be indexed and queried directly if needed. Also makes it explicit from the URL on what page you are which helps for accessibility

blocks/blog-cards/blog-cards.js
```diff
@@ -0,0 +1,119 @@
...
+
+  ul.innerHTML = newUl.innerHTML;
+  window.scrollTo({
+    top: 0,
+    behavior: 'smooth',
+  });
+}
+
+export default function decorate(block) {
+  const observer = new IntersectionObserver(async (entries) => {
```
ramboz, line 105: I would probably start fetching immediately, and we can keep the observer for the rendering if you want, though as it is, it should theoretically be on the initial viewport (above the fold) on most pages anyway, so not sure this has a lot of effect

icons/next.svg
ramboz, line 1: As proposed above, I'd make this a generic icon using `stroke="currentcolor"` and then style if via CSS using color/background/mask as needed

icons/previous.svg
ramboz, line 1: Same here

styles/styles.css
```diff
@@ -347,6 +347,137 @@ main .section.well .default-content-wrapper > * {
   max-width: 1000px;
 }
 
+
+.orange {
```
ramboz, line 351: You probably want to re-use what @travrob1 did in https://github.com/hlxsites/petplace/pull/83#discussion_r1195811917 so you don't have to define all those color classes

blocks/blog-cards/blog-cards.css
varunmitra: This is needed to avoid CLS, I have added an additional property value for 900 px viewport setting min-height to 0

PR #107, opened 2023-05-24
fix: improve index pages template

ramboz:

Fix #<gh-issue-id>

Test URLs:
- Before: https://main--petplace--hlxsites.hlx.page/article/breed/
- After: https://index-pages--petplace--hlxsites.hlx.page/article/breed/


ramboz:
@varunmitra This will set the overall grid layout on the page and inject the pagination block at the end of the grid automatically.
@travrob1 this will also automatically inject the type filters for the breed pages and handle the sidebar decoration with the toggle like we had on the category index pages

PR #111, opened 2023-05-24
Issue 98

### LINKED ISSUE #98: [QA] Home Page
- [x] Slider images appear to be lower resolution than those used on the live site. Is this temporary or how they will look when rolled to production?

- [x] Galaxy Fold: Blog titles, dates, and author names cut off in Category sections.

Image

travrob1:
Paws Up (Articles LIked) Block
Fix #98 

This works is for the initial styles and some minor logic.  A follow up story 
https://github.com/orgs/hlxsites/projects/5/views/1?pane=issue&itemId=29068272
will be for doing that work once we have clearer details from customer.

Test URLs:
- Before: https://main--petplace--hlxsites.hlx.page/article/general/pet-care/best-ways-to-make-a-pet-first-aid-kit
- After: https://issue-98--petplace--hlxsites.hlx.page/article/general/pet-care/best-ways-to-make-a-pet-first-aid-kit

image


blocks/paws-up/paws-up.css
```diff
@@ -1 +1,143 @@
...
+
+.counter .icon svg {
+    color: var(--highlight-color)
+}
+
+.like-post-container p {
+    margin: 15px 0 0;
+    font-family: var(--body-font-family);
+    font-size: .75rem;
+    color: #ff7d5a;
```
ramboz, line 37:
```suggestion
    color: var(--highlight-color);
```

icons/paw-big-filled.svg
ramboz, line 1: Is this really different from the plain `paw.svg`? Could we re-use the same svg and just change it's color?

blocks/paws-up/paws-up.js
```diff
@@ -1,3 +1,41 @@
...
+      <div class="counter">
+        <span class="icon icon-paw" ></span>
+        <p>${count} ${count === 1 ? 'paw' : 'paws'} up</p>
+      </div>
+      <div class="paw-btn-container">
+        <span>Was this article helpful?</span>
+        <span class="paw-btn btn-yes">Yes</span>
+        <span class="paw-btn btn-no">No</span>
+      </div>
+      <img class="email-pets" src="/images/cat-dog.png" alt="email">
```
ramboz, line 25: Ideally we should make a picture tag out of it and serve both the original and an optimized webp version.

blocks/paws-up/paws-up.js
```diff
@@ -1,3 +1,41 @@
...
+      <div class="sign-up-wrapper">
```
ramboz, line 26: I'm wondering if we should start a `/fragments` folder with plain documents in it for such elements so the authors can still go in and update easily

travrob1: I dont see this html as a re-usable fragment.  I have started developing this way the past several days because i think it is easier to have more of an html feel ...rather than using the dom api to create all the elements.  I think a /fragments makes sense for re-usable html blocks.

icons/paw-big-filled.svg
travrob1: I really hated to do this... but i saw no other way.  There is a subtle css transition when you hover over the paw.  Since we are using the svg as `background: url(picture.svg` ... rather than inline icons... the only way to apply the needed color is to us the `mask` css property.... The issue with that is when we are transiting between hover and hover off.  There is a second when the mask is gone... but the background of the entire div is fading from gray to white... and it looks terrible.  I thought it would not be too bad to have an extra svg element.

PR #121, opened 2023-05-26
Issue 29

### LINKED ISSUE #29: feat: cards block

varunmitra:

Fix #29 
Test URLs:
- Before: https://main--petplace--hlxsites.hlx.page/article/category/pet-care/cat-care/
- After: https://issue-29--petplace--hlxsites.hlx.page/article/category/pet-care/cat-care/


blocks/cards/cards.css
```diff
@@ -29,3 +29,120 @@
   aspect-ratio: 4 / 3;
   object-fit: cover;
 }
+
+main {
+  --offst-left: 20px;
```
ramboz, line 34: You have a typo, should be `--offset-left`. I'd also typically scope to the block instead of globally to limit side-effects and conflicts

blocks/cards/cards.css
```diff
@@ -29,3 +29,120 @@
...
+}
+
+
+main a {
+    color: var(--color-gray-dark);
```
ramboz, line 39: This should probably go into the category index template rather than the block

blocks/cards/cards.css
```diff
@@ -29,3 +29,120 @@
...
+main {
+  --offst-left: 20px;
+}
+
+
+main a {
+    color: var(--color-gray-dark);    
+}
+
+@media (min-width: 400px) {
```
ramboz, line 42: Might want to remove the media query otherwise <400px will be unstyled, unless that is what you wanted?

blocks/cards/cards.css
```diff
@@ -29,3 +29,120 @@
...
+    aspect-ratio: 4 / 3;
+    max-width: 100%;
+    object-fit: cover
+  }
+  
+  .cards.article .blogs-card-body > h3{
+    font-size: var(--body-font-size-l);
+    margin-bottom: 10px;    
+    font-family: var(--heading-font-family);
+    font-weight: 800;
```
ramboz, line 90: font weights are typically inherited from the font face definition, so likely not needed here

styles/styles.css
```diff
@@ -50,7 +50,8 @@
   --color-green-transparent: rgb(120 166 141 / 60%);
   --color-magenta-transparent: rgb(169 77 143 / 60%);
   --color-mauve-transparent: rgb(106 55 90 / 60%);
-
+  --color-light-blue: #c4d1e9;
```
ramboz, line 53: You can move it up with the other plain colors, and I'd also rename to `--color-blue-light` for consistency with the previous ones

blocks/cards/cards.js
```diff
@@ -1,18 +1,135 @@
...
+    if (item.Category.toLowerCase() === post.category.toLowerCase()) {
+      category = item.Category;
+      categorypath = item.Path;
+      categorycolor = item.Color;
+    }
+  });
+  postCard.innerHTML = `
+      <div class="blogs-card-image">
+        <a href="${post.path}">${createOptimizedPicture(post.image, `Teaser image for ${post.title}`, eager).outerHTML}</a>
+        <a class="blogs-card-category ${categorycolor}" href="${categorypath}">${category}</a>
```
ramboz, line 30: Instead of defining individual classes for each color, you could follow the approach used in https://github.com/hlxsites/petplace/blob/22cb4ac1bb52971562dc38a7aae3f76544906784/scripts/scripts.js#L468-L470

Essentially you define 3 variables for background, border and text color, and override those at runtime as needed via the `setProperty`. The colors from the categories should already be all mapped as color variables.

styles/styles.css
```diff
@@ -515,3 +516,133 @@ a.sr-focusable:focus {
 }
 
 /* Template overrides */
+
+.orange {
```
ramboz, line 520: All those could be dynamically set via `setProperty` approach mentioned above. It makes it easier to extend later for the devs/authors

icons/previous.svg
travrob1, line 1: two icons is actually not necessary... I built out the previous and next style with just css using to rotate and change color.

https://github.com/hlxsites/petplace/pull/101/files#diff-83dbe39d37d7cb3c974032622b7e872cec1c4c253870021bfab25060df9121c2R183-R196

scripts/ffetch.js
travrob1, line 1: At first glance this ffetch.js script seems a bit overkill.  I'm curious what problem it is solving exactly?

styles/styles.css
travrob1: Yeah i think you might have gotten this css from a commit I made... however i ended up changing directions and not using these explicit classes.... Rather we are using the color variables at the top of the style sheet... and specific variable overrides as seen herej
css:
https://github.com/hlxsites/petplace/blob/4dd5787a119e89e8bb129c60082b1a3efdbabf6b/templates/category-index/category-index.css#L21-L26
js override: 
https://github.com/hlxsites/petplace/blob/fc7ca21a68bf178a365468c81ce02d6537bf0326/templates/category-index/category-index.js#L68

travrob1 (dismissed):

blocks/cards/cards.css
varunmitra: Removed from the main and moved this to the element class

scripts/ffetch.js
varunmitra: check this [slack](https://cq-dev.slack.com/archives/C9KD0TT6G/p1682613378821069) conversation

PR #127, opened 2023-05-30
Issue 43 - Popular Posts Block

### LINKED ISSUE #43: feat: popular posts block
Image

travrob1:
image
image

Fetching popular posts from a document... fallback to home page slideshow.

Fix #43 

Test URLs:
- Before: https://main--petplace--hlxsites.hlx.page/article/dogs/pet-health/hip-dysplasia-in-dogs
- After: https://issue-43--petplace--hlxsites.hlx.page/article/dogs/pet-health/hip-dysplasia-in-dogs

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -17,7 +17,6 @@ module.exports = {
     'import/extensions': ['error', {
       js: 'always',
     }],
-    'max-len': ['error', { code: 999 }],
     'no-return-await': 'off',
   },
 };
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -17,6 +17,7 @@ module.exports = {
     'import/extensions': ['error', {
       js: 'always',
     }],
-    'max-length': 'off',
+    'max-len': ['error', { code: 999 }],
+    'no-return-await': 'off',
   },
 };
--- a/blocks/popular-articles/popular-articles.css
+++ b/blocks/popular-articles/popular-articles.css
@@ -1,1 +1,63 @@
-/* Nothing here yet */
+.popular-articles-container {
+    background-color: var(--background-color-light);
+}
+
+.popular-articles-container h2{
+    text-align: center;
+    margin-top: -20px;
+}
+
+.popular-cards-wrapper {
+    display: flex;
+    justify-content: space-between;
+    flex-direction: column;
+}
+
+.popular-cards-wrapper a {
+    display: flex;
+    margin-bottom: 20px;
+}
+
+.img-div {
+    border-radius: 6px;
+    min-width: 120px;
+    min-height: 120px;
+    max-height: 120px;
+    position: relative;
+    overflow: hidden;
+    margin-right: 20px;
+}
+
+.img-div img{
+    position: absolute;
+    top: 0;
+    left: 0;
+    width: 100%;
+    height: 100%;
+    object-fit: cover;
+    object-position: center center;
+    opacity: 1;
+    transition: opacity 500ms ease 0s;
+
+}
+
+.title-div {
+    color: var(--text-color);
+    margin-right: 10px;
+}
+
+@media (min-width: 768px) {
+    .article-page .section {
+        padding-left: 20px;
+        padding-right: 20px;
+    }
+
+    .popular-cards-wrapper {
+        flex-direction: row;
+    }
+
+    .popular-cards-wrapper a {
+        display: flex;
+        max-width: 33.3%;
+    }
+}
--- a/blocks/popular-articles/popular-articles.css
+++ b/blocks/popular-articles/popular-articles.css
@@ -7,6 +7,10 @@
     margin-top: -20px;
 }
 
+.popular-posts-wrapper {
+    display: flex;
+}
+
 .popular-cards-wrapper {
     display: flex;
     justify-content: space-between;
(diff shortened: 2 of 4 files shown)


.eslintrc.js
```diff
@@ -17,6 +17,7 @@ module.exports = {
     'import/extensions': ['error', {
       js: 'always',
     }],
-    'max-length': 'off',
+    'max-len': ['error', { code: 999 }],
```
ramboz, line 20: Not a big fan of changing the rules unless they really come in the way 😉

blocks/popular-articles/popular-articles.css
```diff
@@ -1 +1,81 @@
...
+    justify-content: space-between;
+    flex-direction: column;
+}
+
+.popular-cards-wrapper a {
+    display: flex;
+    margin-bottom: 20px;
+}
+
+.img-div {
```
ramboz, line 26: I would scope to the block at a minimum to avoid possible conflicts

blocks/popular-articles/popular-articles.js
```diff
@@ -1,3 +1,89 @@
...
+      image: tempElement.querySelector('meta[property="og:image"]').content,
+      imageAlt: tempElement.querySelector('meta[property="og:image:alt"]').content,
+      path,
+      title: tempElement.querySelector('h1').textContent,
+      category: catData.Category,
+      categoryPath: catData.Path,
+    };
+  });
+
+  return await Promise.all(PromiseArray);
```
ramboz, line 26: you can just return the Promise without awaiting here, since you already await the result on L43

blocks/popular-articles/popular-articles.js
```diff
@@ -1,3 +1,89 @@
...
+  if (!popularPosts) {
+    return;
+  }
+  const paths = [...popularPosts.children].map((child) => new URL(child.textContent.trim()).pathname);
+  // eslint-disable-next-line consistent-return
+  return await fetchArticleData(paths);
+}
+
+async function getTopPostsFromSlideshow() {
+  const res = await fetch('/');
```
ramboz, line 47: Could we just fetch `/index.plain.html` to have less stuff to parse? or do you need stuff from the HTML head?

blocks/popular-articles/popular-articles.js
```diff
@@ -1,3 +1,89 @@
...
+  // Get the content within the <main> tag
+  const slideShow = tempElement.querySelector('.slideshow');
+  const paths = [...slideShow.children].map((child) => new URL(child.querySelector('a').href).pathname);
+  paths.splice(3, paths.length);
+
+  return await fetchArticleData(paths);
+}
+
+export default async function decorate(block) {
+  block.innerHTML = '<h2>Popular Posts</h2>';
```
ramboz, line 63: I think we should ideally take this from the `/popular-posts` page directly so authors can change it if needed.

blocks/popular-articles/popular-articles.js
```diff
@@ -1,3 +1,89 @@
...
+  let PopularPostsData = await getPopularPosts();
```
ramboz, line 64: I would propose changing this slightly so we always fetch `/popular-posts`. If it does contain `<a>` elements, we use the top 3 to populate the block, and we otherwise pad with whatever is in the homepage slideshow

blocks/popular-articles/popular-articles.js
travrob1: i tried using /index.plain.html and ran into issues.

blocks/popular-articles/popular-articles.js
travrob1: done 😎

blocks/popular-articles/popular-articles.js
```diff
@@ -1,3 +1,93 @@
...
+    // eslint-disable-next-line max-len
+    paths = [...popularPostsElem.children].map((child) => new URL(child.textContent.trim()).pathname);
+  }
+  // if popularPostsElem is not found or there is less than three paths
+  if (paths.length !== 3) {
+    paths.push(...await getPathsFromSlideshow());
+    paths.splice(3, paths.length);
+  }
+
+  return await fetchArticleData(paths);
```
ramboz, line 67: You can just return, you don't need to await here. the caller will take care of it already

blocks/popular-articles/popular-articles.js
```diff
@@ -1,3 +1,93 @@
...
+
+  if (popularPostsElem) {
+    // eslint-disable-next-line max-len
+    paths = [...popularPostsElem.children].map((child) => new URL(child.textContent.trim()).pathname);
+  }
+  // if popularPostsElem is not found or there is less than three paths
+  if (paths.length !== 3) {
+    paths.push(...await getPathsFromSlideshow());
+    paths.splice(3, paths.length);
+  }
```
ramboz, line 65: I'd suggest adjusting to, so that if the author adds more than 3 rows, we truncate whatever overflows
```suggestion
  if (paths.length < 3) {
    paths.push(...await getPathsFromSlideshow());
  }
  paths.splice(3, paths.length);
```

blocks/popular-articles/popular-articles.js
```diff
@@ -1,3 +1,93 @@
...
+  const PopularPostsData = await getPopularPosts(block);
+
+  const cardWrapper = document.createElement('div');
+  cardWrapper.classList.add('popular-cards-wrapper');
+
+  PopularPostsData.forEach((post) => {
+    const popularPostsWrapper = `
+      <div class="popular-posts-card">
+        <a href=" ${post.path}">
+            <div class="img-div">${createOptimizedPicture(post.image, post.imageAlt).outerHTML}</div>
```
ramboz, line 80: I'd add a size constraint so we download fewer bytes. Also, replace `outerHTML` with an actual `append`:
```js
cardWrapper.innerHTML += popularPostsWrapper;
cardWrapper.querySelector('.img-div').append(createOptimizedPicture(…));
```
`outerHTML` is notoriously really bad from a performance point of view

ramboz (commented):
Left a few comments, and also saw that the heading size on larger screens is smaller than on the prod website. And I do love that you didn't bring over the ellipsis truncation, we do have enough space for the full titles 😄

PR #130, opened 2023-05-31
Issue 118

### LINKED ISSUE #118: feat: add images for each category
Currently, the categories only contain textual data, but the various category pages have different banners each.
So we need to include those in the XSLX sheet as well if possible, and we then need to inject the banner dynamically in the category index template

travrob1:
Fetching custom category hero image
image

Fix #118 

Test URLs:

Default - https://issue-118--petplace--hlxsites.hlx.page/article/category/american-shorthair/ 
Custom - https://issue-118--petplace--hlxsites.hlx.page/article/category/keeping-your-cat-healthy/
Customer - https://issue-118--petplace--hlxsites.hlx.page/article/category/keeping-your-dog-healthy/

### DIFF
--- a/scripts/scripts.js
+++ b/scripts/scripts.js
@@ -92,6 +92,7 @@ export async function getCategoryImage(path) {
   div.innerHTML = htmlText;
 
   const column = div.querySelector('.columns');
+  // eslint-disable-next-line max-len
   return [...column.children].find((el) => el.children[0].textContent.trim() === path)?.children[1].children[0];
 }
 /**
--- a/scripts/scripts.js
+++ b/scripts/scripts.js
@@ -97,6 +97,7 @@ export async function getCategoryImage(path) {
   let matchingPicture;
   for (let i = 0; i < rows.length; i += 1) {
     if (rows[i].children[0].textContent.trim() === path) {
+      // eslint-disable-next-line prefer-destructuring
       matchingPicture = rows[i].children[1].children[0];
       break;
     }
--- a/scripts/scripts.js
+++ b/scripts/scripts.js
@@ -81,6 +81,31 @@ export async function getCategoryByName(categoryName) {
 }
 
 /**
+ * Queries the colum and finds the matching image else uses default image.
+ * @param path
+ * @returns {Promise<HTMLPictureElement>}
+ */
+export async function getCategoryImage(path) {
+  const res = await fetch('/article/category/category-images');
+  const htmlText = await res.text();
+  const html = document.createElement('html');
+  html.innerHTML = htmlText;
+
+  const column = html.querySelector('.columns');
+  const defaultPicture = column.querySelector('picture');
+
+  const rows = [...column.children];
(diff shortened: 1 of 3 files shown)


scripts/scripts.js
```diff
@@ -80,6 +80,31 @@ export async function getCategoryByName(categoryName) {
   return categories.data.find((c) => c.Category === categoryName);
 }
 
+/**
+ * Queries the colum and finds the matching image else uses default image.
+ * @param path
+ * @returns {Promise<HTMLPictureElement || undefined>}
+ */
+export async function getCategoryImage(path) {
+  const res = await fetch('/article/category/category-images');
```
ramboz, line 89: Could we use the `.plain.html` version?

scripts/scripts.js
```diff
@@ -80,6 +80,31 @@ export async function getCategoryByName(categoryName) {
...
+  const htmlText = await res.text();
+  const html = document.createElement('html');
```
ramboz, line 91: *nitpicking*: This isn't technically creating a valid `html` document, but just a regular node that is named `html` (APIs would be different). So to avoid the confusion, it is probably cleaner to just do a `div` instead, or you could also use:
```js
const parser = new DOMParser();
const svg = window.atob(img.src.split(',')[1]);
const html = parser.parseFromString(htmlText, 'text/html');
```
which is what the `innerHTML` would use internally

scripts/scripts.js
```diff
@@ -80,6 +80,31 @@ export async function getCategoryByName(categoryName) {
...
+  let matchingPicture;
+  for (let i = 0; i < rows.length; i += 1) {
+    if (rows[i].children[0].textContent.trim() === path) {
+      // eslint-disable-next-line prefer-destructuring
+      matchingPicture = rows[i].children[1].children[0];
+      break;
+    }
+  }
+
+  return matchingPicture;
```
ramboz, line 106: Could be simplified to:
```js
return [...column.children]
  .find((el) => el.children[0].textContent.trim() === path)
  ?.children[1].children[0];
```

templates/category-index/category-index.js
```diff
@@ -111,6 +111,11 @@ export async function loadEager(main) {
   updateMetadata();
   main.insertBefore(buildSidebar(), main.querySelector(':scope > div:nth-of-type(2)'));
   createTemplateBlock(main, 'pagination');
+  // eslint-disable-next-line no-restricted-globals
+  const heroImg = await getCategoryImage(location.pathname);
+  if (heroImg) {
+    main.querySelector('picture').outerHTML = heroImg.outerHTML;
```
ramboz, line 117: Rather use:
```suggestion
    main.querySelector('picture').replaceWith(heroImg);
```
to avoid the costly transformation and parsing to the HTML string (`outerHTML` is bad 😉)

scripts/scripts.js
travrob1: 👍

PR #140, opened 2023-06-05
feat: add accessibility to slideshow

### LINKED ISSUE #88: feat: make slideshow fully WCAG 2.1 AA compliant
We should make sure the block is fully compliant with WCAG 2.1 AA. We can essentially follow an approach like https://github.com/hlxsites/delta/blob/main/us/en/skymiles/blocks/tabs/aria-tabs.js to create a wrapper web component that abstracts the aria properties and default behavior, and make sure to follow the spec from https://www.w3.org/WAI/ARIA/apg/patterns/carousel/ to do most of the decoration.

To validate the compliance, we can run this through:
- https://wave.webaim.org/ or https://chrome.google.com/webstore/detail/wave-evaluation-tool/jbbplnpkjmmeebjpijfedlgcdilocofh
- https://accessibe.com/accessscan
- https://chrome.google.com/webstore/detail/axe-devtools-web-accessib/lhdoppojpmngadmnindnejefpokejbdd

mfrisbey:
Fix #88

Test URLs:
- After: https://issue-88--petplace--hlxsites.hlx.page/

This PR introduces a new method that will decorate an _existing_ slideshow block with attributes and functionality to meet accessibility requirements. Some of the things provided are:

* Keyboard interaction support for slide selection tabs, if supported by the slideshow block.
* Auto rotation capabilities, if desired by the block.
* Full accessibility-related attribute decoration for all elements of a slideshow, including the slideshow's overall container, its tab selection elements (if provided), and the slides themselves.
* The decorator doesn't do the work of actually visibly changing the active slide, but it will decorate active/inactive slides appropriately, and provide a custom event that consumers can use to do that work.

### DIFF
--- a/blocks/slideshow/aria-slideshow.js
+++ b/blocks/slideshow/aria-slideshow.js
@@ -27,17 +27,19 @@ class ResumableInterval {
     return !this.inputs.mouse && !this.inputs.keyboard;
   }
 
-  pause(inputs = {}) {
+  pause(slideshow, inputs = {}) {
     this.setInputs(inputs);
+    slideshow.setAttribute('aria-live', 'polite');
     clearInterval(this.interval);
   }
 
-  resume(inputs = {}) {
+  resume(slideshow, inputs = {}) {
     this.setInputs(inputs);
     clearInterval(this.interval);
     // ensure that neither mouse nor keyboard has paused
     // the interval
     if (this.areInputsCleared()) {
+      slideshow.setAttribute('aria-live', 'off');
       this.interval = setInterval(() => {
         this.callback();
       }, this.intervalTime);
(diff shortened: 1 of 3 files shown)


blocks/slideshow/aria-slideshow.js
```diff
@@ -0,0 +1,343 @@
...
+ * @param {number} currentIndex Currently activated slide in the slideshow.
+ * @param {number} newIndex Index of the new slide to activate.
+ */
+export function changeSlide(slideshowInfo, currentIndex, newIndex) {
+  if (currentIndex === newIndex) {
+    // optimize if user clicks on the same tab
+    return;
+  }
+  const {
+    slideshow,
```
travrob1, line 87: i think it would be helpful here if `slideshow` was renamed to `slideshowElem`, to add some context as to what slideshow is.

travrob1 (approved):

PR #141, opened 2023-06-05
updating the breadcrumbs markup to be WCAG compliant

travrob1:
Updating markup and visuals for breadcrumbs

Fix #104 

Test URLs:
- Before: https://main--petplace--hlxsites.hlx.page/article/cats/pet-care/cat-care/senior-cat-care/heres-older-cat-wants-attention-often
- After: https://issue-104--petplace--hlxsites.hlx.page/article/cats/pet-care/cat-care/senior-cat-care/heres-older-cat-wants-attention-often

### DIFF
--- a/scripts/scripts.js
+++ b/scripts/scripts.js
@@ -463,41 +463,51 @@ export function initializeTouch(block, slideWrapper) {
  * @returns {Promise<Element>} Resolves with the crumb element.
  */
 export async function createBreadCrumbs(crumbData) {
-  const breadcrumbContainer = document.createElement('div');
-  // Use the last item in the list's color
   const { color } = crumbData[crumbData.length - 1];
+  const breadcrumbContainer = document.createElement('nav');
+  breadcrumbContainer.setAttribute('aria-label', 'Breadcrumb');
 
+  const ol = document.createElement('ol');
+  ol.setAttribute('role', 'list');
+  ol.setAttribute('aria-breadcrumb', 'true');
+
+  const homeLi = document.createElement('li');
+  homeLi.setAttribute('role', 'listitem');
   const homeLink = document.createElement('a');
-  homeLink.classList.add('home');
   homeLink.href = '/';
   homeLink.innerHTML = '<span class="icon icon-home"></span>';
   homeLink.setAttribute('aria-label', 'Go to our Homepage');
-  breadcrumbContainer.append(homeLink);
+  homeLink.setAttribute('aria-current', 'page');
+  homeLi.append(homeLink);
+  ol.append(homeLi);
 
   crumbData.forEach((crumb, i) => {
+    const li = document.createElement('li');
+    li.setAttribute('role', 'listitem');
     if (i > 0) {
       const chevron = document.createElement('span');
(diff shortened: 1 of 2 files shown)


scripts/scripts.js
```diff
@@ -463,42 +463,52 @@ export function initializeTouch(block, slideWrapper) {
...
  */
 export async function createBreadCrumbs(crumbData) {
-  const breadcrumbContainer = document.createElement('div');
-  // Use the last item in the list's color
   const { color } = crumbData[crumbData.length - 1];
+  const breadcrumbContainer = document.createElement('nav');
+  breadcrumbContainer.setAttribute('aria-label', 'Breadcrumb');
 
+  const ol = document.createElement('ol');
+  ol.setAttribute('role', 'list');
```
ramboz, line 471: An `ol` already has `role="list"` behind the scene, so you don't need to add it.

scripts/scripts.js
```diff
@@ -463,42 +463,52 @@ export function initializeTouch(block, slideWrapper) {
...
+  ol.setAttribute('aria-breadcrumb', 'true');
```
ramboz, line 472: I don't think this is a valid aria prop

scripts/scripts.js
```diff
@@ -463,42 +463,52 @@ export function initializeTouch(block, slideWrapper) {
...
+
+  const homeLi = document.createElement('li');
+  homeLi.setAttribute('role', 'listitem');
```
ramboz, line 475: Same as `ol`, the `li` tag already implies `role="listitem"`, so yu can skip this

scripts/scripts.js
```diff
@@ -463,42 +463,52 @@ export function initializeTouch(block, slideWrapper) {
...
   const homeLink = document.createElement('a');
-  homeLink.classList.add('home');
   homeLink.href = '/';
   homeLink.innerHTML = '<span class="icon icon-home"></span>';
   homeLink.setAttribute('aria-label', 'Go to our Homepage');
-  breadcrumbContainer.append(homeLink);
+  homeLink.setAttribute('aria-current', 'page');
```
ramboz, line 480: This property should only be set on the last link in the breadcrumb that actually represents the current page.
See https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-current#values for more details.

scripts/scripts.js
```diff
@@ -463,42 +463,52 @@ export function initializeTouch(block, slideWrapper) {
...
+  homeLi.append(homeLink);
+  ol.append(homeLi);
 
   crumbData.forEach((crumb, i) => {
+    const li = document.createElement('li');
+    li.setAttribute('role', 'listitem');
```
ramboz, line 486: Can be skipped

ramboz:
Might just wanna look at the CLS on the article page. Not sure if we just have flaky tests or maybe something that is a side-effect from the changes

scripts/scripts.js
travrob1: Sure aint ;) Good catch.

PR #145, opened 2023-06-05
feat: add support for hidden descriptive text

### LINKED ISSUE #125: feat: automatically decorate hidden descriptive text
It is usually required for a11y and SEO to have descriptive links and labels. When these are not part of the design, we essentially have to hide this text from the screen but leave it in the markup for assistive technologies and bots to parse.

So we need to add some generic decoration as part of the auto-block logic that will take any `a` and rewraps any `[…]` it contains in a `.sr-only` span.

For instance:
```html
<a href="…">Read more [about The Best Gifts for Dog Moms]</a>
```
needs to end up as:
```html
<a href="…">
  Read more
  <span class="sr-only">about The Best Gifts for Dog Moms</span>
</a>
```

mfrisbey:
Fix #125 

Test URLs:
- After: https://issue-125--petplace--hlxsites.hlx.page/

As describe in the ticket, add new method that will create hidden descriptive text for screen readers. As an example, see the "Traveling with a Pet" slide's "Read More" button. The button itself still says "Read More", but upon inspecting the button there will now be a span whose class is `sr-only` and whose content is "about Traveling with a Pet".

### DIFF
--- a/scripts/scripts.js
+++ b/scripts/scripts.js
@@ -283,6 +283,33 @@ async function buildAutoBlocks(main) {
 }
 
 /**
+ * Creates hidden screen reader content for improving the website's screen
+ * reader compatibility. The method will look for all "a" tags in the given
+ * container, and replace content between square brackets with a span whose
+ * class is "sr-only". For example, the following element:
+ *
+ * <a href="...">Read more [about The Best Gifts for Dog Moms]</a>
+ *
+ * Will be replaced with:
+ *
+ * <a href="...">
+ *  Read more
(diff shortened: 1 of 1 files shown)


ramboz (approved):

travrob1 (approved):

mfrisbey:
@travrob1 My understanding is that if a someone is using a screen reader on the site, a button that only says "Read More" isn't useful without being able to see the image for the corresponding article. So this would add an extra, visually hidden, element that the screen reader could pick up to add more context to the button. i.e. instead of the screen reader reporting a button that just says "Read More", it would report a button that says "Read More about traveling with a pet".

PR #147, opened 2023-06-08
feat: add a11y "skip to" links

### LINKED ISSUE #48: feat: a11y "skip to" links
Image
Image
Image

mfrisbey:
Fix #48 

Test URLs:
- After: https://issue-48--petplace--hlxsites.hlx.page/

This PR adds "skip to" buttons to the site. The buttons are hidden by default, but are revealed when they're focused. They're the first focusable elements on the page, so pressing the tab button will reveal them. Interacting with the buttons will focus the corresponding location in the page:

* "Skip to Content" will focus the `<main>` element.
* "Skip to Menu" will focus the first `li`'s `<a>` element in the page's primary navigation.
* "Skip to Footer" will focus the `<footer>` element.

All of the "Skip to" buttons are in a containing `<nav>` element, which is fixed, covers the entire content of the window, has a transparent background, and a `z-index` of -1. When anything in the `nav` is focused, it's brought to the front with a transparent black background; this is what provides the slightly darker "overlay" of the entire page whenever a "skip to" button is visible.

### DIFF
--- a/scripts/scripts.js
+++ b/scripts/scripts.js
@@ -325,7 +325,7 @@ function createA11yQuickNav(links = []) {
     const button = document.createElement('button');
     button.setAttribute('aria-label', l.label);
     button.href = `#${l.id}`;
-    button.innerHTML = `<span>${l.label}</span><span><span class="a11y-quicknav-button"><span class="a11y-quicknav-symbol">↵</span><span>ENTER</span></span></span>`;
+    button.innerHTML = `${l.label}`;
     button.addEventListener('click', (ev) => {
       ev.preventDefault();
       const el = document.getElementById(ev.currentTarget.href.split('#')[1]);
--- a/scripts/scripts.js
+++ b/scripts/scripts.js
@@ -338,7 +338,7 @@ function createA11yQuickNav(links = []) {
     const button = document.createElement('button');
     button.setAttribute('aria-label', l.label);
     button.href = `#${l.id}`;
-    button.innerHTML = `${l.label}`;
+    button.innerHTML = l.label;
     button.addEventListener('click', (ev) => {
       ev.preventDefault();
       const el = document.getElementById(ev.currentTarget.href.split('#')[1]);
--- a/scripts/scripts.js
+++ b/scripts/scripts.js
@@ -317,6 +317,27 @@ export function decorateScreenReaderOnly(container) {
     });
 }
 
+function createA11yQuickNav(links = []) {
+  const nav = document.createElement('nav');
+  nav.setAttribute('aria-label', 'Skip to specific locations on the page');
+  nav.classList.add('a11y-quicknav', 'sr-focusable');
+  links.forEach((l) => {
+    const button = document.createElement('button');
+    button.setAttribute('aria-label', l.label);
+    button.href = `#${l.id}`;
+    button.innerHTML = `<span>${l.label}</span><span><span class="a11y-quicknav-button"><span class="a11y-quicknav-symbol">↵</span><span>ENTER</span></span></span>`;
+    button.addEventListener('click', (ev) => {
+      ev.preventDefault();
+      const el = document.getElementById(ev.currentTarget.href.split('#')[1]);
+      el.setAttribute('tabindex', 0);
+      el.focus();
+      el.addEventListener('focusout', () => { el.setAttribute('tabindex', -1); }, { once: true });
+    });
+    nav.append(button);
+  });
+  document.body.prepend(nav);
+}
+
 /**
  * Decorates the main element.
  * @param {Element} main The main element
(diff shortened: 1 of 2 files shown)


styles/styles.css
```diff
@@ -465,7 +465,77 @@ a.sr-focusable:focus {
...
+}
+
+.a11y-quicknav > button {
+  top: 40px;
+  z-index: 1000;
+  transition: top .22s ease;
+  opacity: 0;
+}
+
+.a11y-quicknav > button, .a11y-quicknav > button:focus {
```
mfrisbey, line 492: Any suggestions for how to fix a lint error on this line? The error is:

```
Expected selector ".a11y-quicknav > button" to come before selector "button:disabled:hover"
```

But if I move `button:disabled:hover` below this, it complains that `button:disabled:hover` should come before `.a11y-quicknav > button:focus` (which is part of the same rule). I'd have to separate the two into their own rules and duplicate the styles so that I can put `button:disabled:hover` between them.

travrob1: if re-ordering is impossible because there is not way to satasify the error... then i would just silence

 /* stylelint-disable-next-line no-descending-specificity */

styles/styles.css
```diff
@@ -465,7 +465,77 @@ a.sr-focusable:focus {
...
+  border: solid 3px rgb(19 110 248 / 80%);
+  outline: solid 0 #639af9;
+  box-shadow: 0 0 0 5px rgb(19 110 248 / 30%);
+  display: flex;
+  gap: 2rem;
+  align-items: center;
+  justify-content: center;
+  text-shadow: 0 0 #27272d;
+  text-decoration: none;
+  color: #1f2533;
```
travrob1, line 507: Could `var(--color-blue) be used here?  It seems closeish.

styles/styles.css
```diff
@@ -465,7 +465,77 @@ a.sr-focusable:focus {
...
+  font-size: var(--body-font-size-default);
+  transform: none;
+  text-transform: unset;
+}
 
+.a11y-quicknav > button:focus {
+  opacity: 1;
+  top: 20px;
+  z-index: 1001;
```
travrob1, line 516: I would recommend 100..  It is much higher than anything we have used... but not too high to not be overwritten by modal overlays etc...

travrob1 (commented):
Nice work.

scripts/scripts.js
```diff
@@ -317,12 +317,34 @@ export function decorateScreenReaderOnly(container) {
...
 
+function createA11yQuickNav(links = []) {
+  const nav = document.createElement('nav');
+  nav.setAttribute('aria-label', 'Skip to specific locations on the page');
+  nav.classList.add('a11y-quicknav', 'sr-focusable');
+  links.forEach((l) => {
+    const button = document.createElement('button');
+    button.setAttribute('aria-label', l.label);
+    button.href = `#${l.id}`;
+    button.innerHTML = `<span>${l.label}</span><span><span class="a11y-quicknav-button"><span class="a11y-quicknav-symbol">↵</span><span>ENTER</span></span></span>`;
```
ramboz, line 328: I wonder if we could do some of this via CSS `::after` and just set the label here. The rest seems mostly presentational

styles/styles.css
ramboz: Agreed, I never liked that rule to begin with as it focuses on the complexity of the selectors first, and ignores any grouping/scoping concepts that may feel more intuitive for an author to organize the code properly.

styles/styles.css
```diff
@@ -465,7 +465,77 @@ a.sr-focusable:focus {
...
+  background-color: var(--background-color);
+  line-height: 1;
+  border: solid 3px rgb(19 110 248 / 80%);
+  outline: solid 0 #639af9;
+  box-shadow: 0 0 0 5px rgb(19 110 248 / 30%);
+  display: flex;
+  gap: 2rem;
+  align-items: center;
+  justify-content: center;
+  text-shadow: 0 0 #27272d;
```
ramboz, line 505: I'd be tempted to just drop this. it visually doesn't really make a difference

styles/styles.css
ramboz: I'd rather suggest just `var(--text-color)`.

styles/styles.css
```diff
@@ -465,7 +465,77 @@ a.sr-focusable:focus {
...
+  text-transform: unset;
+}
 
+.a11y-quicknav > button:focus {
+  opacity: 1;
+  top: 20px;
+  z-index: 1001;
+}
+
+.a11y-quicknav > button .a11y-quicknav-button {
```
ramboz, line 519: We could probably move all of this to the `::after` and have its content set to `↵ ENTER` via CSS

mfrisbey:
@ramboz @travrob1 I made changes based on your feedback. Using `::after`, in particular, really helped simplify the markup. Feel free to take another look if you'd like, otherwise I'll merge the PR tomorrow.

scripts/scripts.js
```diff
@@ -317,12 +317,34 @@ export function decorateScreenReaderOnly(container) {
...
 
+function createA11yQuickNav(links = []) {
+  const nav = document.createElement('nav');
+  nav.setAttribute('aria-label', 'Skip to specific locations on the page');
+  nav.classList.add('a11y-quicknav', 'sr-focusable');
+  links.forEach((l) => {
+    const button = document.createElement('button');
+    button.setAttribute('aria-label', l.label);
+    button.href = `#${l.id}`;
+    button.innerHTML = `${l.label}`;
```
ramboz, line 328: Could be simplified to:
```suggestion
    button.innerHTML = l.label;
```

styles/styles.css
```diff
@@ -465,7 +465,67 @@ a.sr-focusable:focus {
...
 }
 
+.a11y-quicknav {
+  position: fixed;
+  margin: 0;
+  padding: 30px 0 0 20px;
+  left: 0;
+  top: 0;
+  bottom: 0;
+  right: 0;
```
ramboz, line 475: I think you can use the shorthand notation for this:
```suggestion
  inset: 0;
```

styles/styles.css
```diff
@@ -465,7 +465,67 @@ a.sr-focusable:focus {
...
+
+/* stylelint-disable-next-line no-descending-specificity */
+.a11y-quicknav > button, .a11y-quicknav > button:focus {
+  position: absolute;
+  padding: 18px 20px;
+  border-radius: 50px;
+  background-color: var(--background-color);
+  line-height: 1;
+  border: solid 3px rgb(19 110 248 / 80%);
+  outline: solid 0 #639af9;
```
ramboz, line 501: I wonder if this isn't already inherited from the `:focus-visible` (~L120)

ramboz (approved):

PR #148, opened 2023-06-09
Issue 86

### LINKED ISSUE #85: feat: verify accessibility for the accordion block
Since the accordion block is already leveraging the `aria-accordion.js`, we should have pretty strong compliance to begin with.

Still, we can run this through:
- https://wave.webaim.org/ or https://chrome.google.com/webstore/detail/wave-evaluation-tool/jbbplnpkjmmeebjpijfedlgcdilocofh
- https://accessibe.com/accessscan
- https://chrome.google.com/webstore/detail/axe-devtools-web-accessib/lhdoppojpmngadmnindnejefpokejbdd

### LINKED ISSUE #86: feat: make care tabs fully WCAG 2.1 AA compliant
We should make sure the block is fully compliant with WCAG 2.1 AA. We can leverage https://github.com/hlxsites/delta/blob/main/us/en/skymiles/blocks/tabs/aria-tabs.js to do most of the decoration.

To validate the compliance, we can run this through:
- https://wave.webaim.org/ or https://chrome.google.com/webstore/detail/wave-evaluation-tool/jbbplnpkjmmeebjpijfedlgcdilocofh
- https://accessibe.com/accessscan
- https://chrome.google.com/webstore/detail/axe-devtools-web-accessib/lhdoppojpmngadmnindnejefpokejbdd

travrob1:
Addressing accessibility for Slide Cards, care tabs, and accordion

Fix #85 
Fix #86 

Test URLs:
- After: https://issue-86--petplace--hlxsites.hlx.page/article/breed/pembroke-welsh-corgi


mfrisbey (approved):
Beauty.

blocks/accordion/aria-accordion.js
```diff
@@ -63,15 +63,14 @@ export class AriaAccordion extends HTMLElement {
...
-    [...this.children].forEach((el, i) => {
+    [...this.children].forEach((el) => {
       idBtn = Math.random().toString(32).substring(2);
       idPnl = Math.random().toString(32).substring(2);
 
       const button = document.createElement('button');
       button.id = idBtn;
       button.setAttribute('aria-expanded', false);
       button.setAttribute('aria-controls', idPnl);
-      button.setAttribute('tabindex', i === this.selectedIndex ? 0 : -1);
```
ramboz, line 74: Setting the tab index is actually quite important from an accessibility point of view. You are meant to "tab" into the list, and then navigate inside via arrow keys (not tab).

ramboz (dismissed):
The tabindex is an accessibility requirement, so we don't want to extract it from the aria component. I'm not sure what the issue was with it on mobile, but we probably need to address this differently

blocks/care-tabs/care-tabs.js
```diff
@@ -7,6 +7,21 @@ function setActiveCard(card, details) {
   desktopDetailsContainer.innerHTML = '';
   desktopDetailsContainer.append(details.cloneNode(true));
 }
+
+function setResponsiveAttributes(block, size) {
+  if (size < 1024) {
+    [...block.children].forEach((child) => {
+      const card = child.children[0];
+      card.removeAttribute('tabindex');
```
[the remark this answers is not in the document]
travrob1: Screen Shot 2023-06-12 at 9 34 15 AM
On desktop, you navigate via selecting the cards.
But on mobile you navigate via the buttons.  When the cards have a tabindex... it breaks the block... navigating in a way unintended from the bock... not changing it via the `slide()` method...and thus not keeping track of the slide index and other values correctly.

blocks/accordion/aria-accordion.js
ramboz: Yeah… that phrasing is definitely not ideal. What is meant is that when you are on the accordion header and you hit the `Tab` key, it should move focus to the next focusable element inside the accordion panel (if opened), or outside the accordion container. The thing is that the accordion headers are not considered focusable, just the active one is.

It's slightly better phrased on the tabs: https://www.w3.org/WAI/ARIA/apg/patterns/tabs/
The accordion, per se, doesn't have it's own ARIA role, so we typically fall back to the same as tabs to have consistent behavior

blocks/care-tabs/care-tabs.js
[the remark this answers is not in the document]
ramboz: From an accessibility point of view, I'd say we ideally should have the prev/next buttons `aria-controls` the `tablist` role, and essentially modify the tabindex. I think https://github.com/hlxsites/delta/blob/main/us/en/skymiles/blocks/tabs/aria-tabs.js would already support this.
I also have a hard time understanding how the tabindex alone would conceptually break navigation. I would imagine we have a conflict maybe with the animations, but that shouldn't take precedence over accessibility, so we'd need to fix it there instead.

blocks/care-tabs/aria-care-tabs.js
ramboz, line 1: Ideally, the idea was to have the generic patterns in those aria files, so `aria-tabs` would have been ideal. Anything that is specific to the `care` tabs would be moved to `care-tabs.js`

ramboz (approved):
Technically, looks good!

Ideally, the idea was to have the generic patterns in the aria file so we can re-use the "widget" as-is across projects, and then just have the project-specific code in the block js file. But let's defer that to a later point

PR #150, opened 2023-06-09
integrate google ads into website

### LINKED ISSUE #136: feat: add Google Ads
PetPlace uses Google Publishers Tags (GPT) to place ads on the website.
The current implementation is obsolete:
```html
<script data-ad-client="ca-pub-6577733472295351" async="" src="https://pagead2.googlesyndication.com/pagead/js/adsbygoogle.js" data-checked-head="true"></script>
…
<ins class="adsbygoogle adsbygoogle-noablate" data-adsbygoogle-status="done" style="display: none !important;" data-ad-status="unfilled"></ins>
```

We need to be updated to the latest API versions and best-practices:

https://developers.google.com/publisher-tag/guides/ad-best-practices#use_single_request_architecture_correctly

Petplace is also re-designing how and where they will place ads, but we can already infer that we will need:
- An add placement in the sidebar for articles, like we already have on https://www.petplace.com/article/general/just-for-fun/valentines-from-your-pet/, probably even 2 slots on top of each-other. Those would be auto-decorated as part of the template.
- Inline adds inside articles that authors would need to be able to configure. So we are looking at an `Ads` block where the author can specify an identifier to be targeted by GPT.
- We will likely need a set of predefined block templates in the block library for the inline ads

mfrisbey:

Fix #136 

Test URLs:
- After:
  - https://issue-136--petplace--hlxsites.hlx.page/
  - https://issue-136-lazy--petplace--hlxsites.hlx.page/
  - https://issue-136-eager--petplace--hlxsites.hlx.page/

This PR introduces Google Ads into the site.

### Authoring

Here is how the ads will work from an authoring perspective:

* There is a new `ads` spreadsheet at the root of the project. This file defines the ads that can be used throughout the site, including the ad's path and size. Ads are referenced using their ID.

Screenshot 2023-06-21 at 4 34 16 PM

* Ads can be added to authored pages using the "Ad" block, referencing the desired ad's ID.

Example block

* Which ad appears in the sidebar (or inline for mobile) of an article is determined by its category. There is a new "Ad" column in the categories spreadsheet that should specify the ID of the ad to display. If nothing is specified, the ad will default to `article-default-rail`.

Screenshot 2023-06-21 at 4 35 19 PM

### Technical Details

From an implementation perspective, here is some information:

* When the ad block is decorated, it will take the inner text of its first child element and add it as a `data` attribute to the element. All other children will be removed.
* The article template's lazy load phase will query its category to see if there is an ad defined. If not it will follow its category parent chain until it does find an ad. If no ad is defined anywhere in its chain, it will use the default ID. The template will use this ID to flesh out its ad autoblock in the rail.
* During the main lazy load phase, the page will load the external Google Ad script if there are any ad blocks on the page. It will look up ads by their ID, then provide the information to the Ad API to display the actual ads.

### Examples

Articles that will display an ad:

- https://issue-136--petplace--hlxsites.hlx.page/article/dogs/just-for-fun/the-best-gifts-for-dog-moms
- https://issue-136--petplace--hlxsites.hlx.page/article/cats/pet-care/fall-inspired-cat-names
- https://issue-136-lazy--petplace--hlxsites.hlx.page/article/dogs/just-for-fun/the-best-gifts-for-dog-moms
- https://issue-136-lazy--petplace--hlxsites.hlx.page/article/cats/pet-care/fall-inspired-cat-names
- https://issue-136-eager--petplace--hlxsites.hlx.page/article/dogs/just-for-fun/the-best-gifts-for-dog-moms
- https://issue-136-eager--petplace--hlxsites.hlx.page/article/cats/pet-care/fall-inspired-cat-names


ramboz (commented):
The overall logic looks good!

I'll have to spend some more time to check if we can generalize the ads config retrieval and caching with the categories maybe so we re-use the code, and also how we can reduce the CLS and significant TBT to bump up the score 🤔

mfrisbey:
Thanks for the review, @ramboz! 

> how we can reduce the CLS

I can reduce the current CLS to some degree by increasing the default size of the ad slot. This wouldn't completely eliminate the issue, since we don't know beforehand how big the ad should be. We can fake it for now, though. I'll make that change.

> check if we can generalize the ads config retrieval and caching with the categories maybe so we re-use the code

I can definitely generalize the code more and make it reusable between the two cases (I didn't like needing to do so much copy/paste, anyway). That wouldn't solve the speed issues, but will at least make the code more maintainable.

blocks/ad/ad.js
```diff
@@ -1,3 +1,15 @@
...
+ */
+export default function decorate(block) {
+  if (!block.id) {
+    block.id = `ad-${Math.random().toString(32).substring(2)}`;
+  }
+  [...block.children].forEach((row, index) => {
+    if (index === 0) {
+      block.dataset.adid = String(row.innerText).trim();
+    }
+    row.remove();
```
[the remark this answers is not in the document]
mfrisbey: True, we could do it by adding multiple rows to the same block, or by adding multiple blocks.

blocks/ad/ad.css
```diff
@@ -1 +1,49 @@
...
+  position: relative;
+  height: 100%;
+  min-height: inherit;
+  background-repeat: no-repeat;
+  background-image:
+    radial-gradient(circle 16px, white 99%, transparent 0),
+    radial-gradient(circle 16px, white 99%, transparent 0),
+    linear-gradient(var(--background-color) 32px, transparent 0),
+    linear-gradient(var(--background-color) 32px, transparent 0),
+    linear-gradient(var(--color-gray-light-alt) 100%, transparent 0);
```
ramboz, line 16: We could just keep the "last" one that defines the background. The others are really just there to mimic an expected content structure (which we had for cards), but we don't necessarily need this on ads… since they'll be kinda random

scripts/scripts.js
```diff
@@ -490,6 +506,136 @@ export function addFavIcon(href) {
...
+
+    // Enable SRA and services.
+    window.googletag.pubads().enableSingleRequest();
+    window.googletag.enableServices();
+  });
+  adData.forEach((currAdData) => {
+    window.googletag.cmd.push(() => {
+      window.googletag.display(currAdData.ad.id);
+    });
+  });
```
ramboz, line 635: As part of https://github.com/hlxsites/petplace/tree/issue-136b, I had tried playing around with different approaches to see if we can do anything to help with the TBT. One thing I did was move the slot definition and display directly into the ad block.

Having the ad data in the context of the block rendering also let me set an initial width/height directly in the right context. https://github.com/hlxsites/petplace/blob/issue-136b/blocks/ad/ad.js#L65-L66. I think this might help with the sizing.

On https://issue-136--petplace--hlxsites.hlx.page/ for instance, we initially have a 1344x250 div, when the size in the `ads.xslx` is defined as 970x250. That would fix it.

styles/styles.css
```diff
@@ -858,6 +858,14 @@ a.sr-focusable:focus {
...
 }
 
+main .section.search-container.ad-container .ad-wrapper .ad {
+  height: 100%;
+}
+
+main .section.search-container.ad-container .ad-wrapper {
+  min-height: 250px;
+}
+
```
ramboz, line 868: With the approach above, we'd also likely not need the CSS rules to size the component

ramboz (commented):
Looks good! Left a few ideas as comments

PR #164, opened 2023-06-15
feat: implement /traveling-with-a-pet

### LINKED ISSUE #133: feat: implement /traveling-with-a-pet
The goal here is to migrate https://www.petplace.com/traveling-with-a-pet/ to Franklin and re-implement the template and blocks for the page.

# Scope

## Links:

- Word document in Sharepoint for the page: [traveling-with-a-pet.docx](https://adobe.sharepoint.com/:w:/r/sites/HelixProjects/_layouts/15/Doc.aspx?sourcedoc=%7B61065362-F3D1-4E2C-8277-D4D98A455990%7D&file=traveling-with-a-pet.docx&action=default&mobileredirect=true)
- Page Url: https://main--petplace--hlxsites.hlx.page/drafts/xfeng/traveling-with-a-pet

## Tasks

- [x] #139
- [x] Identify the individual sections and blocks required to design the page
  - The Table-of-Content links at the top
  - The types icon bar
  - etc.
- [x] Identify what existing blocks we can re-use and create variants for
- [x] Create a new template JS file for the page in `/templates` following the model of the existing ones
- [x] Implement individual blocks as needed in `/blocks`

This is meant as a wrapper story and we can extract individual bite-size tickets from this as needed.

# Screenshots

## Desktop

Image
Image
Image

ramboz:

Fix #133 

Test URLs:
- Before: https://main--petplace--hlxsites.hlx.page/traveling-with-a-pet
- After: https://issue-133--petplace--hlxsites.hlx.page/drafts/xfeng/traveling-with-a-pet


FentPams:
### **Desktop**
**Hero image and [Anchor](https://github.com/hlxsites/petplace/tree/issue-133/blocks/anchor):** 
Screenshot 2023-06-19 at 9 53 20 AM
**[float-column](https://github.com/hlxsites/petplace/tree/issue-133/blocks/float-column)**
Screenshot 2023-06-19 at 9 56 16 AM
**[icon-type](https://github.com/hlxsites/petplace/tree/issue-133/blocks/icon-type)**
Screenshot 2023-06-19 at 9 56 23 AM
**city-section ([link-box](https://github.com/hlxsites/petplace/tree/issue-133/blocks/link-box), [tips](https://github.com/hlxsites/petplace/tree/issue-133/blocks/tips))**
Screenshot 2023-06-19 at 9 56 30 AM
Screenshot 2023-06-19 at 9 56 37 AM

### **Tablet**
Screenshot 2023-06-19 at 10 04 59 AM
Screenshot 2023-06-19 at 10 05 06 AM
Screenshot 2023-06-19 at 10 05 12 AM
Screenshot 2023-06-19 at 10 05 21 AM
Screenshot 2023-06-19 at 10 05 28 AM

### **Mobile**
Screenshot 2023-06-19 at 10 06 33 AM
Screenshot 2023-06-19 at 10 06 42 AM
Screenshot 2023-06-19 at 10 06 48 AM
Screenshot 2023-06-19 at 10 07 15 AM
Screenshot 2023-06-19 at 10 07 23 AM

blocks/float-column/float-column.js
```diff
@@ -0,0 +1,23 @@
+import { decorateResponsiveImages } from '../../scripts/scripts.js';
+
+const isPicture = (block) => !!block.querySelector('picture');
+
+export default function decorate(block) {
+  // for each block, decorate image and text according to their positions
+  const floatItem = block.querySelectorAll('div[data-valign="middle"]');
```
ramboz, line 7: I don't think `data-valign` would be reliably set in the markup. This will be dependent on the cell styling in authoring, so it's likely not going to work for all use cases you have

blocks/float-column/float-column.js
```diff
@@ -0,0 +1,23 @@
...
+  floatItem.forEach((item, index) => {
+    if (isPicture(item)) {
+      const direction = index % 2 === 0 ? 'left' : 'right';
+      item.classList.add(`float-column-image-${direction}`);
+      decorateResponsiveImages(item, [600, 900]);
+    } else {
+      item.classList.add('text');
+    }
+
+    if (index % 2 === 1 && isPicture(item)) {
```
ramboz, line 18: Instead of testing twice `isPicture(item)`, you could store the result and reuse it

blocks/icon-type/icon-type.css
```diff
@@ -0,0 +1,97 @@
...
+}
+
+.icon-type-list li icon-item {
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+  width: 100%;
+  height: 100%;
+  border-right: 1px solid #ff7d5a;
```
ramboz, line 30:
```suggestion
  border-right: 1px solid var(--highlight-color);
```

blocks/icon-type/icon-type.css
```diff
@@ -0,0 +1,97 @@
...
+  border-right: none;
+}
+
+.icon-type-list li icon-item img {
+  width: 100%;
+  height: 65%;
+}
+
+.icon-type-list span {
+  color: #736aa4;
```
ramboz, line 43: This is likely close enough to `--color-purple` that we could re-use it

blocks/icon-type/icon-type.js
```diff
@@ -0,0 +1,49 @@
...
+
+  // Find the li elements containing the desired text
+  const liElements = iconList.querySelectorAll('li');
+
+  // Iterate over each li element
+  liElements.forEach((liElement) => {
+    const textContent = liElement.textContent.trim();
+    const a = transformTextContent(textContent);
+    // Fetch the SVG file from the 'icons' folder
+    const svgFilePath = `/icons/${a}.svg`;
```
ramboz, line 23: For icons, you can directly use `:iconName:` in the document and Franklin will automatically transform this to an icon reference for you (so `:home:`, `:logo:`, `:city-hiking:`). You can also leverage `toClassName` to transform a string to a standardized string (i.e. `toClassname('City Luxury')` gives us `city-luxury`)

blocks/tips/tips.js
```diff
@@ -0,0 +1,18 @@
...
+  const path = '/icons/tips-star.svg';
+  const alt = 'Tips Star Image';
+
+  divElement.innerHTML = `<img src=${path} alt=${alt}>`;
+
+  // Get the parent element.
+  const parentElement = block.querySelector('.tips.block > div');
+
+  // Insert the 'tips-icon' at the beginning of the parent element.
+  parentElement.insertBefore(divElement, parentElement.firstChild);
```
ramboz, line 17: We could probably achieve the same using pure CSS and just a `::before` element that has the image as a background image or on the `content` property..

icons/city-beach.svg
ramboz, line 1: I would suggest editing all those SVG files and replace all the explicit `#736AA4` color reference to plain `currentcolor` value. You can then style the color via CSS by setting `.icon { color: #736AA4; }`.

That way the icon is easier to style dynamically based on the color scheme

templates/category-index/category-index.css
ramboz, line 1: Can you revert the changes in this file? This isn't part of the page you are working on and it could introduce undesired side-effects

templates/traveling-page/traveling-page.css
```diff
@@ -0,0 +1,347 @@
+/*Style hero image*/
+.traveling-page .hero {
+  box-sizing: border-box;
+  display: flex;
+  flex-direction: column;
+  justify-content: end;
+  height: 62.5vw;
+  min-height: 0;
```
ramboz, line 8: This is the default, so shouldn't be needed.

templates/traveling-page/traveling-page.css
```diff
@@ -0,0 +1,347 @@
...
+.section.city h3 {
+  font-size: calc(1.325rem + .9vw);
+  line-height: 1.1;
+  color: var(--background-color-dark);
+  margin-bottom: 1.875rem;
+  font-weight: 800;
+  background-color: var(--background-color);
+}
+
+.city .icon-type-list {
```
ramboz, line 92: I'd move this back into the icon type block's CSS, and potentially add a block variant if needed to differentiate from the usage at the top of the page.

Maybe `Icon Type` vs. `Icon Type (inline)`

templates/traveling-page/traveling-page.css
```diff
@@ -0,0 +1,347 @@
...
+}
+
+.city .icon-type-list li {
+  padding-left: 0.25em;
+  margin-bottom: 1rem;
+  height: 1rem;
+  flex: 0 0 10rem;
+}
+
+.city .icon-type-list icon-item {
```
ramboz, line 109: These `icon-item` would likely be replaced by `.icon` once you address the comments above.

templates/traveling-page/traveling-page.css
```diff
@@ -0,0 +1,347 @@
...
+  display: flex;
+  flex-direction: row;
+  justify-content: left;
+  width: 100%;
+  height: 150%;
+  border-right: 1px solid #ff7d5a;
```
ramboz, line 115:
```suggestion
  border-right: 1px solid var(--highlight-color);
```

templates/traveling-page/traveling-page.css
```diff
@@ -0,0 +1,347 @@
...
+    grid-template-areas:
+      "name name name name name name name name name name"
+      "icons icons icons icons icons icons icons icons icons icons"
+      "links links links links links links links links links links"
+      "img1 img1 img1 img1 img1 img1 img1 img1 img1 img1"
+      "txt1 txt1 txt1 txt1 txt1 txt1 txt1 txt1 txt1 txt1"
+      "txt2 txt2 txt2 txt2 txt2 txt2 txt2 txt2 txt2 txt2 "
+      "img2 img2 img2 img2 txt3 txt3 txt3 txt3 txt3 txt3"
+      "img3 img3 img3 img3 txt3 txt3 txt3 txt3 txt3 txt3"
+      "tips tips tips tips tips tips tips tips tips tips";
```
ramboz, line 277: Could we simplify to?
```suggestion
      "name name"
      "icons icons"
      "links links"
      "img1 img1"
      "txt1 txt1"
      "txt2 txt2"
      "img2 txt3"
      "img3 txt3"
      "tips tips";
```
and potentially play with the column sizes as needed

templates/traveling-page/traveling-page.css
```diff
@@ -0,0 +1,347 @@
...
+    padding-bottom: 4.5rem;
+    column-gap: 4rem;
+    grid-template-areas:
+      "img1 img1 img1 img1 img1 name name name name name name name"
+      "img1 img1 img1 img1 img1 icons icons icons icons icons icons icons"
+      "img1 img1 img1 img1 img1 links links links links links links links"
+      "img1 img1 img1 img1 img1 txt1 txt1 txt1 txt1 txt1 txt1 txt1"
+      "txt2 txt2 txt2 txt2 txt2 txt2 txt2 txt2 img2 img2 img2 img2"
+      "img3 img3 img3 img3 txt3 txt3 txt3 txt3 txt3 txt3 txt3 txt3"
+      "img3 img3 img3 img3 tips tips tips tips tips tips tips tips";
```
ramboz, line 316: I'd simplify and try to align to a halves/thirds/quarters layout but not more granular than that. It's just too many columns to manage

templates/traveling-page/traveling-page.js
```diff
@@ -0,0 +1,54 @@
+export async function loadEager() {
+  const links = document.querySelectorAll('a');
+  links.forEach((link) => link.setAttribute('target', '_blank'));
+  return false;
```
ramboz, line 4: Not needed

templates/traveling-page/traveling-page.js
```diff
@@ -0,0 +1,54 @@
+export async function loadEager() {
+  const links = document.querySelectorAll('a');
+  links.forEach((link) => link.setAttribute('target', '_blank'));
```
ramboz, line 3: I don't think it's a good idea to convert all links on the page to opening in a new window/tab. At best this would be a "block-level" decision.

Maybe what you are actually looking for is to open external links in new tabs?

templates/traveling-page/traveling-page.js
```diff
@@ -0,0 +1,54 @@
...
+    const elements = [
+      '.default-content-wrapper',
+      '.icon-type-wrapper',
+      '.link-box-wrapper',
+      ['.columns.city-header.block>div>div:nth-child(1)', 'city-header-img'],
+      ['.columns.city-header.block>div>div:nth-child(2)', 'city-header-txt'],
+      ['.columns.city-middle.block>div>div:nth-child(1)', 'city-middle-txt'],
+      ['.columns.city-middle.block>div>div:nth-child(2)', 'city-middle-img'],
+      ['.columns.city-footer.block>div>div:nth-child(1)', 'city-footer-img'],
+      ['.columns.city-footer.block>div>div:nth-child(2)', 'city-footer-txt'],
```
ramboz, line 27: Selecting based on index isn't usually ideal, since it forces the authors to follow those indexes precisely to get the layout working. We usually recommend to auto-decorate classes based on what is in the div (like text vs img), rather than based on the index alone.

I would also move the img/txt class addition directly into the block logic. Essentially modifying the columns.
It already sets `columns-img-col`, and you could add `columns-text-col` as an alternative class.

templates/traveling-page/traveling-page.js
```diff
@@ -0,0 +1,54 @@
...
+        const el = section.querySelector(e[0]);
+        el.classList.add(e[1]);
+        return el;
+      }
+      return section.querySelector(e);
+    });
+
+    // rearrange the elements under section city
+    const arrangedCitySection = document.createElement('div');
+    arrangedCitySection.classList.add('section', 'city');
```
ramboz, line 40: I would suggest using the section metadata in the document to do this:
```
| Section Metadata |
|-------+----------+
| Style | City     |
```

templates/traveling-page/traveling-page.js
```diff
@@ -0,0 +1,54 @@
...
+
+    section.replaceWith(arrangedCitySection);
+
+    // pick the best grid area layout by image2 width/height
+    const img2 = elements[6].querySelector('img');
+    const width = img2.getAttribute('width');
+    const height = img2.getAttribute('height');
+    if (width / height > 1.9) {
+      arrangedCitySection.classList.add('backup-layout');
+    }
```
ramboz, line 52: I would suggest moving this into the columns block logic, and probably set a `columns-img-col-vertical` vs. `columns-img-col-vertical` on the column itself, and then a `has-(vertical|horizontal)-image` on the parent section from there as needed.

blocks/anchor/anchor.css
ramboz, line 1: I have extracted our existing "table of content" logic into a dedicated block. I wonder if we could combine the 2 as a single block with 2 variants. You block could become a `ToC (inline)` or similar, and we could combine the JS logic

templates/category-index/category-index.css
ramboz, line 1: I would revert the changes here, they are not part of the scope of the PR

blocks/anchor/anchor.js
```diff
@@ -0,0 +1,35 @@
...
+    const newDiv = document.createElement('div');
+    newDiv.classList.add('anchor-list');
+
+    cities.forEach((city) => {
+      const newAnchor = document.createElement('a');
+      newAnchor.href = `#${city.id}`;
+      newAnchor.innerText = city.innerText;
+      newDiv.appendChild(newAnchor);
+    });
+    const anchor = document.querySelector('.anchor');
```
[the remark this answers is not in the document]
FentPams: You are right. I didn't realize it. Nice catch! I'll also check if other places I make same mistakes. Thank you!

blocks/float-column/float-column.js
FentPams: > Instead of testing twice `isPicture(item)`, you could store the result and reuse it

Yes! it save time on running selector again! Thank you!

templates/traveling-page/traveling-page.js
FentPams: I thought opening link in the same tab will distract users, and make users harder to direct back. Then I read this article, I noticed it is more suitable in our case to open in same tab.  https://uxdesign.cc/linking-to-a-new-tab-vs-same-tab-f88b495d2187. Thank you for this advice.

blocks/icon-type/icon-type.js
FentPams: - Thank you for introducing `toClassName`. Replaced my original function to this. 

1. For the icons, I think the icon-type block in my page not only includes icons, but also with text attached. And probably easier for authors to expect the coming text( what they input in the icon-type column, like `Mountainous` will be directly shown in the page), which I believe is more straightforward. 

But thank you for mentioning that, will handle similar case using this idea.

blocks/tips/tips.js
FentPams: Amazing idea!! Solved. Thank you :)

templates/traveling-page/traveling-page.css
FentPams: Thank you for the idea. Moved into icon block ` Icon Type (inline)`

icons/city-beach.svg
FentPams: Good point! Done.

templates/traveling-page/traveling-page.css
FentPams: It seems like it is not default, if I comment out these lines, the h1 with its background will float to the top (also, the line2 - 10 in category-index.css also declare these.)
Maybe we can extract it and make it a global thing?

templates/traveling-page/traveling-page.js
FentPams: Good point! Done.

templates/traveling-page/traveling-page.css
FentPams: The case here is icon with name, not a pure icon. So, I am afraid `.icon` can't work.

templates/traveling-page/traveling-page.css
FentPams: If you were suggesting the following code, I think because each item in a different row span differently, which means they need to be separately defined. The code becomes lengthy and not straightforward. 

```
img1 {
  grid-column: 1 / span 5;
  grid-row: 1;
}

name {
  grid-column: 6 / span 6;
  grid-row: 1;
}

icons {
  grid-column: 6 / span 6;
  grid-row: 2;
}

links {
  grid-column: 6 / span 6;
  grid-row: 3;
}

txt1 {
  grid-column: 6 / span 6;
  grid-row: 4;
}

txt2 {
  grid-column: 1 / span 5;
  grid-row: 5;
}

img2 {
  grid-column: 9 / span 4;
  grid-row: 5;
}

txt3 {
  grid-column: 5 / span 6;
  grid-row: 6;
}

img3 {
  grid-column: 1 / span 4;
  grid-row: 6;
}

tips {
  grid-column: 5 / span 7;
  grid-row: 7;
}
```


Do you think this will make it more readable and cleaner if I make all the area names aligned?
```   
grid-template-areas:
      "name name name name name name name name name name"
      "tags tags tags tags tags tags tags tags tags tags"
      "link link link link link link link link link link"
      "img1 img1 img1 img1 img1 img1 img1 img1 img1 img1"
      "txt1 txt1 txt1 txt1 txt1 txt1 txt1 txt1 txt1 txt1"
      "txt2 txt2 txt2 txt2 txt2 txt2 txt2 txt2 txt2 txt2 "
      "img2 img2 img2 img2 txt3 txt3 txt3 txt3 txt3 txt3"
      "img3 img3 img3 img3 txt3 txt3 txt3 txt3 txt3 txt3"
      "tips tips tips tips tips tips tips tips tips tips";
```

ramboz: If you look at your layout, the columns do line up properly, so you can simplify to what I suggested above.
Screenshot 2023-06-24 at 8 26 03 AM

You can then use the column sizing, like:
```css
  grid-template:
      "name name"
      "icons icons"
      "links links"
      "img1 img1"
      "txt1 txt1"
      "txt2 txt2"
      "img2 txt3"
      "img3 txt3"
      "tips tips" / 1fr 2fr;
```
to have the grid lay all this out in a 33% / 66% size.

blocks/anchor/anchor.css
FentPams: Done. The current ToC(inline) is only for 'Traveling with a pet' page. We could make it more generic after.

templates/category-index/category-index.css
FentPams: Reverted.

templates/traveling-page/traveling-page.css
FentPams: Thank you. Solved!

templates/traveling-page/traveling-page.js
FentPams: Do you have idea about **ensuring the completion** of decorating blocks **before** load lazy ()?

After I moved img/txt class into block logic, seems like a race condition occurs. The logic of decorating columns is not finished sometimes before loadlazy, leading to the failure of following the flow: 
1. getting each element using the class name we just defined in the columns block, 
2. then assign/decorate them in template. (i.e.`.columns-img-col` class has not yet been added before loadlazy() starts dealing with the first or second city sections in 6* slower CPU case)

My possible solutions and problem:

- Moving all the block assigning and decorating logic into columns, but as my page has a separate `tips` column, it takes the last cell and has layouts like:

Screenshot 2023-06-25 at 10 53 18 AM

That probably means I still need to identify elements from the columns, i.e.` footer-img`, `footer-text`, and delegate the cell they can go to with tips column. Then the same issue comes.

- Put tips inside of the footer paragraph (which I think is not a good choice), make italic of tips or :tips: , then read and handle it inside of columns block.  In this way, the tips will be the logic of the columns block.

Would like to hear your idea. Thank you!

FentPams: I come up with the solution by `setTimeOut`, which can ensure the blocks are decorated before running loadlazy. But the lighthouse score is not pleasant. I guess I will try to hide the city section when they are not done decorated.
Screenshot 2023-06-25 at 6 00 43 PM

ramboz: Since all the blocks are decorated in the lazy phase, there isn't much you can do here before that. Using `await` on the block decoration would at least make sure the block has fully executed before continuing the logic, but the template is still executed before the blocks anyway… so that won't solve your issue.

I was also about to propose to leverage `setTimeout`, but with no timeout (or `0`) so we just push the executing at the end of the stack. Might be enough to have the blocks render first.

Alternatively, we could use a `MutationObserver` on the block that listens for changes to the `class` attribute, and then run the template logic in it. Something like:
```js
const observer = new MutationObserver((entries) => {
  entries.forEach((entry) => {
    if (entry.attributeName !== 'class') {
      return;
    }
    // Do the magic here
  });
});
observer.observe(/* column element */, { attributes: true });
```

PR #170, opened 2023-06-20
feat: add skeleton loading animation to cards loading logic

ramboz:
Please always provide the [GitHub issue(s)](../issues) your PR is for, as well as test URLs where your change can be observed (before and after):

Fix #169

Test URLs:
- Before: https://main--petplace--hlxsites.hlx.page/article/category/pet-care/
- After: https://issue-169--petplace--hlxsites.hlx.page/article/category/pet-care/

PR #179, opened 2023-06-22
issue 49 sort button implementation

### LINKED ISSUE #49: feat: search results page
It seems that BambooHR has some search capabilities that are already implemented in a good way: https://cq-dev.slack.com/archives/C9KD0TT6G/p1683793127084509

We can probably follow a similar path, and would likely need to improve our query indexes

varunmitra:

Fix #49 

Test URLs:
- After: https://issue-49--petplace--hlxsites.hlx.page/search?query=dogs


ramboz (commented):
Overall looks good 👍🏼 

I do see some CLS on desktop though. I also had the issue on category index pages. Maybe you can follow the "skeleton" placeholder approach I had there to limit this:
https://github.com/hlxsites/petplace/pull/170/files#diff-40fdce80e7e6d904fa7553815650489b0930ba3c387052b3e4e1826c4a5004d7R16-R30
You might also be able to get rid of the TBT on mobile using the metering logic in there.

PR #221, opened 2023-07-05
fix: article author should link to author page (if it exists)

### LINKED ISSUE #191: fix: article author should link to author page (if it exists)

FentPams:
Fix #191 

Test URLs:
- Before: https://main--petplace--hlxsites.hlx.live/article/general/pet-care/how-do-pet-microchips-work
- After: https://issue-191--petplace--hlxsites.hlx.live/article/general/pet-care/how-do-pet-microchips-work
Screenshot 2023-07-05 at 10 24 58 AM

### DIFF
--- a/blocks/article-author/article-author.js
+++ b/blocks/article-author/article-author.js
@@ -1,11 +1,15 @@
+import ffetch from '../../scripts/ffetch.js';
 import { decorateIcons, getMetadata } from '../../scripts/lib-franklin.js';
 
-export default function decorate(block) {
+export default async function decorate(block) {
   const author = getMetadata('author');
+  const authorIndex = await ffetch('/authors/author-index.json').filter((item) => item.Name === author).first();
+  const path = authorIndex?.Path;
   const date = getMetadata('publication-date');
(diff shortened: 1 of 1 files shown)


blocks/article-author/article-author.js
```diff
@@ -1,11 +1,15 @@
+import ffetch from '../../scripts/ffetch.js';
 import { decorateIcons, getMetadata } from '../../scripts/lib-franklin.js';
 
-export default function decorate(block) {
+export default async function decorate(block) {
   const author = getMetadata('author');
+  const authorIndex = await ffetch('/authors/author-index.json').filter((item) => item.Name === author).first();
```
ramboz, line 6: you can skip ffetch here since we'll only need the first handful of results (ffetch is good when we need pagination).
```suggestion
  const authorIndex = await fetch('/authors/author-index.json').find((item) => item.Name === author);
```
I would probably also use `toLowerCase()` on the string comparison, or even standardize to `toClassName` so we account for minor input variations

ramboz (approved):
Looks good. Minor suggestions to improve the logic a bit

PR #266, opened 2023-07-21
implement cost of owning a puppy page

### LINKED ISSUE #263: feat: implement https://www.petplace.com/cost-of-owning-a-puppy/
Image
Image
Image
Image
Image

mfrisbey:
Fix #263

Test URLs:
- After:
  - https://issue-263--petplace--hlxsites.hlx.page/cost-of-owning-a-puppy
  - https://issue-263--petplace--hlxsites.hlx.page/cost-of-owning-a-puppy?martech=off

NOTE: this page will require publishing /cost-of-owning-a-puppy.docx and /category/categories.xlsx (the page is added as its own category).

This PR adds two new blocks:

## Price Range

Replaces the SVG image that was on the previous site. When authored, the block will contain simple text representing a price range:

Screenshot 2023-07-21 at 11 13 26 AM

Desktop resolution rendering on site:

Screenshot 2023-07-21 at 11 13 35 AM

Tablet/phone rendering on site:

Screenshot 2023-07-21 at 11 13 43 AM

## Article Disclaimer

Displays "fine print" informing the visitor about the page's monetization. When authored, the block will contain the disclaimer text to display in the article:

Screenshot 2023-07-21 at 11 13 14 AM

On all resolutions, the rendered output will be a small, purple-colored font:

Screenshot 2023-07-21 at 11 13 57 AM

### DIFF
--- a/blocks/article-disclaimer/article-disclaimer.css
+++ b/blocks/article-disclaimer/article-disclaimer.css
@@ -1,0 +1,5 @@
+.article-disclaimer {
+  margin-bottom: 0.75em;
+  font-size: var(--body-font-size-xs);
+  color: var(--color-magenta);
+}
--- a/blocks/article-disclaimer/article-disclaimer.css
+++ b/blocks/article-disclaimer/article-disclaimer.css
@@ -1,5 +1,0 @@
-.article-disclaimer {
-  margin-bottom: 0.75em;
-  font-size: var(--body-font-size-xs);
-  color: var(--color-magenta);
-}
--- a/blocks/gauge/gauge.js
+++ b/blocks/gauge/gauge.js
@@ -1,0 +1,6 @@
+export default function decorate(block) {
+  const divs = block.querySelectorAll('div div');
+  divs[1].classList.add('cell1');
+  divs[2].classList.add('cell2');
+  divs[3].classList.add('cell3');
+}
--- a/blocks/gauge/gauge.js
+++ b/blocks/gauge/gauge.js
@@ -1,6 +1,6 @@
 export default function decorate(block) {
   const divs = block.querySelectorAll('div div');
-  divs[1].classList.add('cell1');
-  divs[2].classList.add('cell2');
-  divs[3].classList.add('cell3');
+  divs[1].classList.add('gauge-icon');
+  divs[2].classList.add('gauge-title');
+  divs[3].classList.add('gauge-value');
 }
--- a/blocks/price-range/price-range.css
+++ b/blocks/price-range/price-range.css
@@ -1,0 +1,147 @@
+.price-range .expenses::before, .price-range .expenses::after, .price-range::before {
+  position: absolute;
+  width: 75px;
+  height: 75px;
+  border-radius: 50%;
+}
+
+.price-range .expenses::before {
+  z-index: 10;
+  left: 119px;
+  top: -48px;
+  content: '$';
+  display: flex;
+  gap: 0;
+  align-items: center;
+  justify-content: center;
+  border: 6.5px solid var(--text-color);
+  border-radius: 50%;
+  font-size: var(--heading-font-size-xl);
+  font-weight: 900;
+  color: var(--background-color);
+}
+
+.price-range::before {
+  content: '';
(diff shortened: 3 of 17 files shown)


mfrisbey:
@ramboz Are you aware of anything that's generally causing performance issues on article pages? I can dig into why this might be loading so slowly, but wanted to make sure there isn't a known issue first.

ramboz:
@mfrisbey `.hlx.page` is typically not using performant caching that the `.hlx.live` would have on the edge.
The consequence is that calls to the `query-index.json` used in the article nav are typically slow and cause TBT and increased SI. On top of that the martech stack (google tag manager, google ads and pushly) also all tend to have more visible impact on the stage environment and further decrease the score.

But you got a pretty bad run in the first place. I'm pretty sure retrying would give us better results

blocks/price-range/price-range.css
```diff
@@ -0,0 +1,139 @@
+.price-range .expenses::before, .price-range .expenses::after, .price-range::before {
```
ramboz, line 1: I'd be tempted to generalize even further. Let's say we call this a `gauge` or `linear meter` block or whatever makes better sense to a native speaker 😉 
We could then design the block in 3 columns: `| icon | label | value |`

blocks/price-range/price-range.js
```diff
@@ -0,0 +1,8 @@
+export default async function decorate(block) {
+  block.querySelector('div').classList.add('range');
+
+  const expenses = document.createElement('div');
+  expenses.classList.add('expenses');
+  expenses.innerText = 'Average Range of Expenses:';
```
ramboz, line 6: With the proposal above, this would move to the second column in the block

blocks/price-range/price-range.css
```diff
@@ -0,0 +1,139 @@
...
+  width: 75px;
+  height: 75px;
+  border-radius: 50%;
+}
+
+.price-range .expenses::before {
+  z-index: 10;
+  left: 119px;
+  top: -48px;
+  content: '$';
```
ramboz, line 12: With the proposal above, this would move to the first column in the block.

ramboz (commented):
Looks great overall!

I'd propose to generalize even more and make this a generic block they could re-use for other purposes as well. Might also be worth adding to the block library

mfrisbey:
> I'd propose to generalize even more

Love it, I'll work on those changes.

blocks/article-disclaimer/article-disclaimer.css
```diff
@@ -0,0 +1,5 @@
+.article-disclaimer {
```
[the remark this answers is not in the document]
mfrisbey: We possibly could. I looked at that block initially, but it's powered by a content fragment whose text is different. Maybe we could add optional content to the block, which would be the URL to the fragment to use? It would default to the currently hard-coded fragment URL. Then we could add a new fragment for this text. WDYT?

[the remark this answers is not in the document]
mfrisbey: Oh, maybe the text _is_ the same :joy: I will just reuse that one.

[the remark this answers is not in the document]
mfrisbey: I keep speaking too soon. It's _slightly_ different. This article's disclosure begins with "The following is advertorial content." The disclosure block doesn't, so if it needs to match exactly, we'd need to come up with a solution.

mfrisbey:
@ramboz I made changes to the PR based on your feedback - would you please take another look?

blocks/gauge/gauge.js
```diff
@@ -0,0 +1,6 @@
+export default function decorate(block) {
+  const divs = block.querySelectorAll('div div');
+  divs[1].classList.add('cell1');
+  divs[2].classList.add('cell2');
+  divs[3].classList.add('cell3');
```
ramboz, line 5: I think we should name those semantically, like `.gauge-icon`, `.gauge-title` and `.gauge-value`

ramboz (approved):
Looks overall good!

Would just suggest renaming the sub-element classes

PR #275, opened 2023-07-26
fix:  previous/next article doesn't exist in its own category

### LINKED ISSUE #270: fix: empty prev/next article navigation
On some pages like https://www.petplace.com/article/dogs/pet-health/6-deadly-poisons-that-could-kill-your-dog, we reach the end of the category and hence there are no next and/or previous articles. In that case, we should plan for a proper fallback.

Options discussed in https://adobe-dx-support.slack.com/archives/C04P5DFPDK9/p1689987199287679

FentPams:
When the previous/next article doesn't exist in its own category, using its parent/sibling category's article's last/first article as a fallback.

Fix #270 
Screenshot 2023-07-28 at 10 33 02 AM
Case 1's example url:
https://issue-270--petplace--hlxsites.hlx.page/article/cats/pet-care/cat-care/senior-cat-care/tip-cats-considered-seniors
Screenshot 2023-07-28 at 10 38 22 AM
Case 2's example url:
https://issue-270--petplace--hlxsites.hlx.page/article/birds/general/selecting-the-best-bird-feeder
Screenshot 2023-07-28 at 10 37 51 AM

Test URLs:
- Before: https://main--petplace--hlxsites.hlx.page/article/dogs/pet-care/160-littermate-names-for-dogs
- After:
  - https://issue-270--petplace--hlxsites.hlx.page/article/dogs/pet-care/160-littermate-names-for-dogs

### DIFF
--- a/blocks/article-navigation/article-navigation.js
+++ b/blocks/article-navigation/article-navigation.js
@@ -50,7 +50,7 @@ function createArticleDetails(block, key, categoryInfo, article) {
   block.append(sectionContainer);
 }
 
-function getFilteredArticles(categories) {
+function getFilteredArticles(category) {
   // Get all articles in that category
   const articles = ffetch('/article/query-index.json')
     .sheet('article')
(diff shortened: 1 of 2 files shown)


ramboz (commented):
The current approach would essentially query and traverse all articles 3 times.

I would suggest refactoring so we only traverse once by modifying the filter to also include parent categories matching on the `path`.
We can then:

1. find the position of the current page in the articles
2. slice the array in 2 before and after that position
3. reverse the "before" array and find the 1st one in the same category, and fall back to the first one in the list if there isn't one
4. find the 1st one in the "after" in the same category, and fall back to the first one in the list if there isn't one

FentPams:
New implementation for the ticket: https://github.com/hlxsites/petplace/pull/284

PR #279, opened 2023-07-27
fix:add margin on ad after tiles, adjust the width of the iframe

### LINKED ISSUE #278: fix : ad margin with tiles and width overflow
### issue1 -- lose margin:
image

### issue2 -- width overflow
Screenshot 2023-07-27 at 8 41 39 AM

FentPams:
Fix #278 
Screenshot 2023-07-27 at 9 48 40 AM
Screenshot 2023-07-27 at 9 49 12 AM

Test URLs:
- After:
  - https://issue-278--petplace--hlxsites.hlx.page/

### DIFF
--- a/templates/home-page/home-page.css
+++ b/templates/home-page/home-page.css
@@ -76,6 +76,10 @@
   flex: 1.5;
 }
 
+.home-page .ad-container.tiles-container .ad-wrapper iframe {
+  max-width: 100%;
+}
+
 @media (min-width: 600px) {
   .home-page .hero h1 {
     font-size: 2.667em;
@@ -113,6 +117,10 @@
   .home-page .ad-container.columns-container .ad-wrapper {
     flex-basis: 25%;
     max-width: 336px;
+  }
+
+  .home-page .ad-container.tiles-container .tiles-wrapper + .ad-wrapper{
+    margin-top: 2rem;
   }
 }
 
--- a/templates/home-page/home-page.css
+++ b/templates/home-page/home-page.css
@@ -76,10 +76,6 @@
   flex: 1.5;
 }
 
-.home-page .ad-container.tiles-container .ad-wrapper iframe {
-  max-width: 100%;
-}
-
 @media (min-width: 600px) {
   .home-page .hero h1 {
     font-size: 2.667em;


FentPams:
@mfrisbey  Hi Mark, I adjusted the width of the ads to avoid overflow, while the ads itself will be cut off when screen shrink, any better idea on doing this? Thank you.

mfrisbey:
@FentPams are you refreshing the page after the screen shrinks? As far as I know, Google selects the ad based on the initial screen size, and doesn't select a different ad as the window is resized.

FentPams:
> @FentPams are you refreshing the page after the screen shrinks? As far as I know, Google selects the ad based on the initial screen size, and doesn't select a different ad as the window is resized.

@mfrisbey  no, I didn't refresh, but shrink the screen size manually, so the ads is not re-selected and cause the width overflow. Does this case we need to handle? 
And, I am also wondering how the ads are loaded, seems not every time of page load, the ads will be loaded.
Thank you

mfrisbey:
> no, I didn't refresh

As long as the overflow is fixed (or the ad disappears) after refreshing the screen, I think that will be enough.

> shrink the screen size manually... this case we need to handle?

IMO, no. We, as developers, are resizing the screen manually a lot because we're testing, but I don't think the average user is going to do this. Especially on mobile devices where the screen size is fixed. It's always good practice to try to handle this, but if we tried to request a different ad every time the `resize` event fires, we would probably exceed our quota for Google's ad API :)

> I am also wondering how the ads are loaded, seems not every time of page load

We depend on the Google ads API for this. I've also noticed that sometimes the ads appear, and sometimes they don't. The API seems very flaky that way. It could also be that we're not using the API correctly (this is the first time Julien or I have worked with Google ads), so maybe it would be good to have another set of eyes look at it, too :)

FentPams:
@mfrisbey  Could you review it and help to merge the code, I only add margin on top of the ad-wrapper, and revert the changes on iframe width. Thank you.
I'll also take a look at the google ads and to see if any possible cause of inconsistent loading.

ramboz:
>  I've also noticed that sometimes the ads appear, and sometimes they don't. The API seems very flaky that way.

@mfrisbey My understanding is that petplace is selling the ad slots to its partners. If there is a bid on the slot, we will have an ad, but if no one payed for it then you won't have anything. Ultimately, as popularity grows, we will likely always have ads 😄

PR #280, opened 2023-07-27
bug: handle possible broken links and log with sampleRUM

### LINKED ISSUE #271: fix: https:///article links lead to broken navigation
Our importer seems to have truncated some of the links to `https:///article/*` and we are essentially missing the domain in there.
We should fix those links automatically during the eager phase and inject the domain in there if possible.

I think we can check the `href` attribute (not the property, since that will actually inject some domain automatically) on the links:
- if it starts with `/`, we are good to go
- if `new URL(…).hostname` does not contain a `.`, we likely have a broken link and should fix the link to be just absolute like the case above
- otherwise link should be good

FentPams:
Fix #271 

Test URLs:
- Before: https://main--petplace--hlxsites.hlx.page/article/dogs/pet-health/dog-health/dog-diet-nutrition/how-much-water-does-a-puppy-need
- After:
  - https://issue-271--petplace--hlxsites.hlx.page/article/dogs/pet-health/dog-health/dog-diet-nutrition/how-much-water-does-a-puppy-need

### DIFF
--- a/scripts/scripts.js
+++ b/scripts/scripts.js
@@ -546,24 +546,23 @@ function fixLinks() {
 
   links.forEach((link) => {
     const href = link.getAttribute('href');
-    // handle href not starts with '/'
-    //   e.g. /article/..... is good link
-    //   e.g. http://article/..... is broken link
+    // handle href not starting with '/'
+    //   e.g. /article/... is a valid link
+    //   e.g. http://article/... is an invalid link
     if (!href.startsWith('/')) {
-      // log the broken link
-      sampleRUM('fix-links-in-article', { source: window.location.href, target: href });
-      /* eslint no-empty: ["error", { "allowEmptyCatch": true }] */
       try {
         const url = new URL(href);
         // If the hostname doesn't contain '.', it's likely broken
         if (!url.hostname.includes('.')) {
+          // log the broken link
+          sampleRUM('fix-links-in-article', { source: window.location.href, target: href });
           // Fix the link by making it absolute
-          // Suppose the common broken link is article/...., the article/ is readed as hostname,
+          // Suppose the common broken link is article/...., the article/ is read as hostname,
           // which should be added back as pathname
           link.setAttribute('href', `${window.location.origin}/${url.hostname}${url.pathname}`);
         }
       } catch (e) {
-
+          // Do nothing here, keep the link as-is
       }
     }
   });
--- a/scripts/scripts.js
+++ b/scripts/scripts.js
@@ -548,24 +548,19 @@ function fixLinks() {
     const href = link.getAttribute('href');
     // handle href not starts with '/'
     //   e.g. /article/..... is good link
-    //   e.g. article/..... is broken link
+    //   e.g. http://article/..... is broken link
     if (!href.startsWith('/')) {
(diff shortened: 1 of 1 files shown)


scripts/scripts.js
```diff
@@ -541,6 +541,36 @@ export async function decorateMain(main) {
...
 }
 
+function fixLinks() {
+  const links = document.querySelectorAll('a');
+
+  links.forEach((link) => {
+    const href = link.getAttribute('href');
+    // handle href not starts with '/'
+    //   e.g. /article/..... is good link
+    //   e.g. article/..... is broken link
```
ramboz, line 551: the second example would actually rather be `e.g. https://article/…`

scripts/scripts.js
```diff
@@ -541,6 +541,36 @@ export async function decorateMain(main) {
...
+      sampleRUM('fix-links-in-article', { source: window.location.href, target: href });
+      try {
+        const url = new URL(href);
+        // If the hostname doesn't contain '.', it's likely broken
+        if (!url.hostname.includes('.')) {
+          // Fix the link by making it absolute
+          // Suppose the common broken link is article/...., the article/ is readed as hostname,
+          // which should be added back as pathname
+          // console.log('before fixing', link.getAttribute('href'));
+          link.setAttribute('href', `https://www.petplace.com/${url.hostname}${url.pathname}`);
```
ramboz, line 563: I'd go with:
```suggestion
          link.setAttribute('href', `${window.location.origin}/${url.hostname}${url.pathname}`);
```
so it dynamically adjustes to hlx.page, hlx.live and prod urls

scripts/scripts.js
```diff
@@ -541,6 +541,36 @@ export async function decorateMain(main) {
...
+          // console.log('fixed link', link.getAttribute('href'));
+        }
+      } catch (e) {
+        // else direct the link to the homepage
+        link.setAttribute('href', 'https://www.petplace.com');
```
ramboz, line 568: I would not necessarily erase the link if it's not the edge case detected, we could be actually removing valid links we haven't thought of.

ramboz:
Looks like a glitch. The script that runs the tests somehow timed out. I re-ran it, and it looks all 🟢 now

FentPams:
> Looks like a glitch. The script that runs the tests somehow timed out. I re-ran it, and it looks all 🟢 now

@ramboz Then good to go if codes look good to you.  Thanks 👍🏻

scripts/scripts.js
```diff
@@ -541,6 +541,31 @@ export async function decorateMain(main) {
...
+
+  links.forEach((link) => {
+    const href = link.getAttribute('href');
+    // handle href not starts with '/'
+    //   e.g. /article/..... is good link
+    //   e.g. http://article/..... is broken link
+    if (!href.startsWith('/')) {
+      // log the broken link
+      sampleRUM('fix-links-in-article', { source: window.location.href, target: href });
+      const url = new URL(href);
```
ramboz, line 555: You'd still need the try/catch around this since URL might be invalid, I would just leave the catch clause empty and do nothing about invalid links.

ramboz (approved):

PR #283, opened 2023-08-02
feat: implement holiday treats article

### LINKED ISSUE #267: feat: implement https://www.petplace.com/holiday-treats-for-dogs/
Image
Image
Image
Image
Image
Image
Image
Image
Image

mfrisbey:
This PR adds a new "Full" hero variant for the "Holiday Treats for Dogs" article. I also modified the disclosure block so that it can accept the URL to a different fragment to display in the disclosure, or to display static content specified in the block.

Lastly, there is a new blade block variant that will add spacing between each item in the blade. The blade is also modified to allow a single cell of content.

If merged, we would need to publish:

* /holiday-treats-for-dogs.docx
* /fragments/amazon-disclosure.docx

Fix #267 

Test URLs:
- Before: https://main--petplace--hlxsites.hlx.page/holiday-treats-for-dogs
- After:
  - https://issue-267--petplace--hlxsites.hlx.page/holiday-treats-for-dogs
  - https://issue-267--petplace--hlxsites.hlx.page/holiday-treats-for-dogs?martech=off


blocks/disclosure/disclosure.js
```diff
@@ -1,9 +1,6 @@
 export default async function decorate(block) {
-  if (block.textContent.trim()) {
-    return;
-  }
-
-  const response = await fetch('/fragments/disclosure.plain.html');
+  const fragment = block.textContent.trim();
+  const response = await fetch(fragment || '/fragments/disclosure.plain.html');
```
ramboz, line 3: I'd even go a step further and do something like:
- if we have text content and it is a valid URL, fetch that
- if we have text content  and it isn't a valid URL, use the text for the content
- otherwise load the default disclosure

templates/full-hero-page/full-hero-page.js
```diff
@@ -0,0 +1,22 @@
+// eslint-disable-next-line import/prefer-default-export
+export function loadLazy() {
```
ramboz, line 2: Seeing that all the logic in this template relates to laying out the hero, I'm thinking we should maybe just explicitly have a hero block in the document and decorate it properly on top of the generic page template

mfrisbey: @ramboz The CSS extends into other areas than just the hero, though. Would we want to add CSS that isn't specific to the hero into the hero block? If so, what would be the best way to do this? Somehow reference the hero's parent in the CSS?

Alternatively, as I said in the original description, I think the page looks pretty good in the article page template, so we could run it by the PetPlace team to see if they want a new page style, or just want to reuse the existing one.

ramboz: The article template is always an option, but I think we are working around the "issue" if we do this in that we have now several root pages that all need a similar layout, but all have slight differences. I think there is probably a way to generalize this into 1 generic template that is just more flexible.

We could combine:
- a basic CSS grid system (something like https://jsfiddle.net/ramboz/367pe0x5/)
- an explicit hero block, instead of the auto-blocking one, so we can have control over variants
- some page + section metadata to have local style variants

I don't have a clear picture yet, so maybe it's best we move forward with this for now and refactor later once we see a clearer pattern emerging

ramboz (approved):

ramboz (approved):

PR #284, opened 2023-08-03
fix: previous/next article doesn't exist in its own category

### LINKED ISSUE #270: fix: empty prev/next article navigation

FentPams:
Re-implement the prev/next, and record the first and last articles from search results. When the previous article is null, use the last article as the previous one. And when the next article is null, use the first article as the next.

Fix #275 

Test URLs:
- After:
  - https://issue-270-1--petplace--hlxsites.hlx.page/
  - https://issue-270-1--petplace--hlxsites.hlx.page/?martech=off

### DIFF
--- a/blocks/article-navigation/article-navigation.js
+++ b/blocks/article-navigation/article-navigation.js
@@ -65,8 +65,8 @@ async function createNavigation(block) {
   if (!categoryInfo) {
     return false;
   }
-  const categories = await getCategoriesPath(categoryInfo.Path);
-
+  // get parent cateogry and all its sub categories(include current category apperantly)
+  const categories = await getCategoriesPath(categoryInfo['Parent Path']);
   // Get all articles in that category
   const articles = ffetch('/article/query-index.json')
     .sheet('article')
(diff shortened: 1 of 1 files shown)


ramboz (commented):
This is definitely cleaner and simpler. Great work!

You'd still have issues if only 1 or 2 articles are in a category. I think we need to adjust the filtering on `ffetch` as well to include also articles in all the parent categories.

FentPams:
> This is definitely cleaner and simpler. Great work!
> 
> You'd still have issues if only 1 or 2 articles are in a category. I think we need to adjust the filtering on `ffetch` as well to include also articles in all the parent categories.

@ramboz I found in that case, if we using `ffetch`, the articles returned back will be the mix from current category and parent category, so that the order of articles in current category cannot be guaranteed, which means even if the article has prev/next in its own category, it is likely will be replaced by the article from parent category, as they are "index closer". What do you think about this?

ramboz:
@FentPams Yes, indeed.

I proposed a solution to this in https://github.com/hlxsites/petplace/pull/275#pullrequestreview-1557947024

At a high level, the idea is to retrieve articles in all the parent categories with ffetch. Then:
1. Save those articles in an array
2. Find the current article in that list, note the position down, and slice the articles array in 2 arrays. 1 before, and 1 after
3. Then iterate over the 2 arrays
  - the array before should have all newer articles (iterate in reverse order so you start from the current article and find the one published right after)
  - the array after should have all the older articles (iterate normally here)
4. When you iterate, note down the immediate previous/next articles (these are the fallbacks), and continue iterating through the list until you find one in the same category as the current article (based on the category in the url)
5. If you find one in each array, then we have a prev/next article in the same category, and if you don't have one, you can fall back to the ones you noted down in step 4.

PR #285, opened 2023-08-04
fix: tracing to all parent categories if prev/next article not exist in own category

FentPams:
Fix #284 #275
Screenshot 2023-08-04 at 2 34 20 AM

Test URLs:
- Before: https://main--petplace--hlxsites.hlx.page/article/small-mammals/pet-care/small-pet-care/small-mammals/petplace-guide-hamster-cages
- After:
  - https://issue-270-2--petplace--hlxsites.hlx.page/article/small-mammals/pet-care/small-pet-care/small-mammals/petplace-guide-hamster-cages
  - https://issue-270-2--petplace--hlxsites.hlx.page/?martech=off/

### DIFF
--- a/blocks/article-navigation/article-navigation.js
+++ b/blocks/article-navigation/article-navigation.js
@@ -50,6 +50,25 @@ function createArticleDetails(block, key, categoryInfo, article) {
   block.append(sectionContainer);
 }
 
+async function getAllParentCategories(category) {
+  const parentCategories = [];
+  let currentCategory = category;
+  while (currentCategory.Parent) {
+    // eslint-disable-next-line no-await-in-loop
+    currentCategory = await getCategory(toClassName(currentCategory.Parent));
+    parentCategories.push(currentCategory);
+  }
+  return parentCategories;
+}
+
+function ifArticleBelongsToCategories(article, categories) {
+  const articleCategories = article.category !== '0'
+    ? article.category.split(',').map((c) => c.trim().toLowerCase())
+    : article.path.split('/').splice(-2, 1);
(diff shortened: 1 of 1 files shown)


FentPams:
@ramboz 
Thanks to your feedback yesterday, I reimplement in two pr:
The implementation in this pr #285 

- Pros: retrieves all parent categories, including indirect parents, up to the top level, which provide a comprehensive view of the category hierarchy.
- Cons:  does not  handle sibling categories, will look up to parent category instead

The implementation in the latest commit in pr #284 

- Pros: considers both the direct parent category and sibling categories
- Cons: focus only on direct parent category, will probably miss relevant articles in indirect parent categories. e.g., if the direct parent category and its own category has limited articles, also no other siblings, will cause the case : the prev/next are all the article itself.

My thought:
I'd recommend this solution#285, as we are more likely to have "the higher parent category is, the more articles are. When we choose to fall back to parent categories and even higher parent categories, would be safer." The edge case I provided above,  article 3-1 has both prev/ next as article b-1 will seldom happen, as its skip parent category B usually have many articles.

ramboz:
@FentPams Agreed. I wouldn't go the siblings' category route. This just increases the dataset we have to traverse and will essentially slow down the page more than the benefit we get out of it

FentPams:
> @FentPams Agreed. I wouldn't go the siblings' category route. This just increases the dataset we have to traverse and will essentially slow down the page more than the benefit we get out of it

@ramboz Glad I got one implementation on the right track😉 let me know if any adjustment needed on the code side, I'll address it.Or it is good to go. Appreciate it:)

ramboz (approved):
Looks good to me

PR #286, opened 2023-08-04
chore: try optimized font loading for smoother UX

### LINKED ISSUE #274: chore: try optimized font loading for smoother UX

ramboz:
Track if fonts have previously been loaded and cached by the browser so we can speed up the loading on subsequent visits. If the browser cache is not disabled, this should remove the flickering from the font swap when you navigate across pages.

This ultimately helps improve LCP, TTI and SI scores.

Fix #274 

Test URLs:
- After:
  - https://issue-274--petplace--hlxsites.hlx.page/
  - https://issue-274--petplace--hlxsites.hlx.page/?martech=off

### DIFF
--- a/scripts/scripts.js
+++ b/scripts/scripts.js
@@ -55,6 +55,18 @@ export function loadScript(url, callback, attributes) {
   return head.querySelector(`script[src="${url}"]`);
 }
 
+/**
+ * Load fonts.css and set a session storage flag
+ */
+async function loadFonts() {
+  await loadCSS(`${window.hlx.codeBasePath}/styles/fonts.css`);
+  try {
+    if (!window.location.hostname.includes('localhost')) sessionStorage.setItem('fonts-loaded', 'true');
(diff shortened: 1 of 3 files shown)


FentPams:
Wow, looks good!! 👍

FentPams:
Might be a dumb question: then why we need a separate fonts css if they will adding toll?

ramboz:
When you switch pages, there is a visual flicker when the font is applied. This is mostly visible on the hero:
Screenshot 2023-08-04 at 8 27 49 AM
Screenshot 2023-08-04 at 8 27 56 AM

With this change, we reduce that effect by loading the fonts definitions sooner once the actual font files are cached. This gives the end user a more stable experience with less shifting. The tradeoff is that we do add another roundtrip to fetch a new CSS file (fonts.css) and this adds 40~50ms delay.

ramboz:
#288  has a better implementation of this that has less impact

PR #292, opened 2023-08-08
feat: implement get-pet-insurance-now page

mfrisbey:
Implements the get-pet-insurance-now page, which consists of some minor margin improvements to the blade block (so that it can support text with multiple paragraphs), and implementing the `pet-insurance-quote` block.

The `pet-insurance-quote` block authoring looks like this:

Screenshot 2023-08-08 at 4 06 35 PM

When rendered, the first cell becomes the label, and the second cell becomes the button:

Screenshot 2023-08-08 at 4 07 28 PM

Screenshot 2023-08-08 at 4 07 50 PM

Fix #291 

Test URLs:
- Before: https://main--petplace--hlxsites.hlx.page/frisbey/drafts/get-pet-insurance-now
- After:
  - https://issue-291--petplace--hlxsites.hlx.page/frisbey/drafts/get-pet-insurance-now
  - https://issue-291--petplace--hlxsites.hlx.page/frisbey/drafts/get-pet-insurance-now?martech=off

PR #296, opened 2023-08-09
feat: implement give-the-gift-of-pet-insurance page

### LINKED ISSUE #293: feat: implement https://www.petplace.com/give-the-gift-of-pet-insurance/
Image
Image
Image
Image

mfrisbey:
The only missing thing for this page is a new variant of the `pet-insurance-quote` block:

Screenshot 2023-08-09 at 10 12 55 AM

Screenshot 2023-08-09 at 10 13 26 AM

It includes additional text, and has different behavior when resizing from mobile to desktop. I created a "triple" variant of the block, but would be open to suggestions if there might be a better way:

Screenshot 2023-08-09 at 10 14 43 AM

The variant requires an additional cell in the block table, and uses a CSS grid to reorder the cells and display them correctly.

Fix #293 

Test URLs:
- Before: https://main--petplace--hlxsites.hlx.page/drafts/frisbey/give-the-gift-of-pet-insurance
- After:
  - https://issue-293--petplace--hlxsites.hlx.page/drafts/frisbey/give-the-gift-of-pet-insurance
  - https://issue-293--petplace--hlxsites.hlx.page/drafts/frisbey/give-the-gift-of-pet-insurance?martech=off


ramboz:
I like that you just created a variant for the block.
I'm not a huge fan of the `triple` term though, as it feels technical. I wonder if we could go with:
- `simple` => what you did in #292
- `full` => what you have here

Another approach could also be to go with different fragments and assume that the customer would create several variants of it depending on context (I do remember a plain image version we used to have as well on the article sidebar).
So we could just name them: `variant-1`, `variant-2`, etc. and they can then extend as needed

mfrisbey:
I agree that the name `full` is better, so I renamed it. For now I'd propose that we do block variants, and if it's looking like we're going to end up with a bunch of variants, or with the block on a lot of different pages, I will explore the content fragment idea.

mfrisbey:
@ramboz Just to clarify, you're meaning that the block would just have one cell with everything on its own line? Something like:

```
+-----------------------------+
| pet insurance quote         |
+-----------------------------+
| Want to learn more...       |
| Check out PetPartners...    |
| Get Your Free Quote         |
+-----------------------------+
```

ramboz:
Yes that would be the general idea

mfrisbey:
@ramboz WDYT of this approach? The two block variants now look like this when authored:

Screenshot 2023-08-10 at 10 22 26 AM

Screenshot 2023-08-10 at 10 22 44 AM

ramboz:
I definitely like the 1st. not sure about the 2nd 🤔

I do feel that having a single cell that we format ourselves for the 2nd kinda leads to an expectation that we control all of its content and layout, while there actually will be some magic arrow injected and the layout will be changed.

I'd be tempted to handle the second as:
```
+------------------------------+
| Pet insurance quote (inline) |
+------------------------------+
| Get Your Free Quote          |
| :right-arrow:                |
| [Get Your Free Quote]()      |
+------------------------------+
```
so we control everything, and the variant becomes just a layout option. So making the icon explicit.

I'd probably also then adjust the classes variant names and have the default behavior being what we have as "full" now, and then have `horizontal`/`inline` as a variant for the other one

ramboz (approved):
This looks good to me!

mfrisbey:
> @ramboz making the icon explicit.

I tried playing with that, but the challenge comes when the block switches to mobile resolutions. When that happens the arrow needs to disappear, so we'd be back to where we are on needing to make assumptions about the content of the block.

> I'd probably also then adjust the classes variant names and have the default behavior being what we have as "full" now, and then have horizontal/inline as a variant for the other one

Switching the default behavior around would be quite a bit of work, so as a compromise I renamed "full" to "vertical."

PR #297, opened 2023-08-10
feat: add newsletter subscription functionality

### LINKED ISSUE #257: feat: newsletter subscription
## Use case

Capture new leads via a newsletter signup form & popup so that they can be targeted in iterable for marketing purposes.

See details in: https://adobe-dx-support.slack.com/archives/C04P5DFPDK9/p1690993241120259

## Technical

- Iterable API Key: `3e7a9624572b4827b156af44e72fceaa`
- Iterable documentation: https://support.iterable.com/hc/en-us/articles/208183076-Field-Data-Types

### Mobile popup
Image

### Desktop popup
Image

### Homepage footer
Image

mfrisbey:
Implements newsletter subscription functionality.

### Newsletter Spreadsheet

Added a new spreadsheet called `newsletter.xlsx` at the root of the project. The content of the newsletter form itself is driven by this spreadsheet.

### Newsletter Signup Block

A new block with two variants. The authoring experience is as follows:

Screenshot 2023-08-22 at 10 55 57 AM

The default variant will maintain a vertical layout. Eventually this block could be used in the pop-ups and on the newsletter signup page. Here is a rendered example at its widest:

Screenshot 2023-08-10 at 4 49 28 PM

The "horizontal" variant will behave as above on smaller resolutions, but will switch to a wider layout on bigger screens. This variant is used in the footer. Here is a rendered example when it's horizontal:

Screenshot 2023-08-10 at 4 16 45 PM

The block is also fully integrated with Iterable. I submitted my own email and received email confirmation for signing up for the newsletter. There is some basic user feedback, showing an error message if the request to Iterable fails, and showing a success message if the request succeeds.

### Newsletter Auto Blocking

Added an autoblocked newsletter sign up form in the `<footer>`. The form displays on all pages:

Screenshot 2023-08-10 at 4 20 16 PM

Article pages have an abbreviated footer:

Screenshot 2023-08-10 at 4 21 04 PM

### TODO

Still need to work on the pop-ups, but I might recommend that as a separate PR to keep things manageable.

Fix #257 

Test URLs:
- After:
  - https://issue-257--petplace--hlxsites.hlx.page/
  - https://issue-257--petplace--hlxsites.hlx.page/?martech=off

### DIFF
--- a/blocks/newsletter-signup/newsletter-signup.js
+++ b/blocks/newsletter-signup/newsletter-signup.js
@@ -59,14 +59,12 @@ export default function decorate(block) {
   const buttonContainer = document.createElement('p');
   buttonContainer.classList.add('button-container');
 
-  let buttonClick = 0;
-
   const button = document.createElement('a');
   button.href = '#';
   button.title = 'Sign up now';
   button.classList.add('button', 'primary');
   button.innerText = 'Sign up now';
-  button.addEventListener('click', (e) => {
+  button.addEventListener('click', async (e) => {
     e.preventDefault();
     enableElement(button, false);
 
@@ -98,18 +96,27 @@ export default function decorate(block) {
       mergeNestedObjects: true,
       createNewFields: true
     };
-    console.log('Calling API with payload', payload);
-    // simulate API call
-    setTimeout(() => {
-      buttonClick += 1;
-      // very other button click will generate an error
-      if (buttonClick % 2 === 1) {
+
+    const fetchOpts = {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+        'Api-Key': '3e7a9624572b4827b156af44e72fceaa'
+      },
+      body: JSON.stringify(payload)
+    };
+    try {
+      const res = await fetch('https://api.iterable.com/api/users/update', fetchOpts);
+      if (!res.ok) {
         showElement(apiError);
         enableElement(button);
       } else {
         showElement(success);
       }
-    }, 1000);
+    } catch (e) {
+      showElement(apiError);
+      enableElement(button);
+    }
   });
   buttonContainer.append(button);
   target.append(buttonContainer);
--- a/blocks/newsletter-signup/newsletter-signup.js
+++ b/blocks/newsletter-signup/newsletter-signup.js
@@ -47,7 +47,7 @@ async function submitForm(block, fd) {
 
 export default async function decorate(block) {
   const form = await createForm(
-    `${window.location.protocol}//${window.location.host}/newsletter.json`,
+    '/newsletter.json',
     (fd) => submitForm(block, fd),
   );
   form.querySelector('label[for="email"]').classList.add('sr-only');
--- a/blocks/newsletter-signup/newsletter-signup.js
+++ b/blocks/newsletter-signup/newsletter-signup.js
@@ -1,60 +1,61 @@
 import { createForm } from '../form/form.js';
 
-function addCheckbox(parent, id, label) {
-  const inputId = `newsletter-checkbox-${id}`;
-  const container = document.createElement('span');
-  container.classList.add(inputId);
-  const check = document.createElement('input');
-  check.setAttribute('type', 'checkbox');
-  check.setAttribute('id', inputId);
-  check.setAttribute('name', inputId);
-  check.setAttribute('value', id);
-  check.checked = true;
-
-  const labelElem = document.createElement('label');
-  labelElem.setAttribute('for', inputId);
-  labelElem.innerText = label;
-
-  container.append(check);
-  container.append(labelElem);
-  parent.append(container);
-
-  return inputId;
+function showMessage(block, message, clazz = 'success') {
+  const messageElement = block.querySelector('.newsletter-message');
+  messageElement.innerText = message;
+  messageElement.classList.remove('success', 'error');
+  messageElement.classList.add(clazz);
 }
 
-function addTextbox(parent, id, label) {
-  const inputId = `newsletter-text-${id}`;
-  const text = document.createElement('input');
-  text.classList.add(inputId);
-  text.setAttribute('type', 'text');
-  text.setAttribute('id', inputId);
-  text.setAttribute('name', inputId);
-  text.setAttribute('placeholder', label);
-  parent.append(text);
-
-  return inputId;
+function showError(block, fd) {
+  showMessage(block, fd.Failure, 'error');
+  block.querySelector('button').removeAttribute('disabled');
 }
 
-function enableElement(target, enable = true) {
-  if (enable) {
-    target.classList.remove('disabled');
-  } else {
-    target.classList.add('disabled');
-  }
-}
-
-function showElement(target, show = true) {
-  if (show) {
-    target.classList.remove('hidden');
-  } else {
-    target.classList.add('hidden');
+async function submitForm(block, fd) {
+  const dognewsletter = document.getElementById('dogs').checked;
+  const catnewsletter = document.getElementById('cats').checked;
+  const email = String(document.getElementById('email').value).trim();
+  const payload = {
+    email,
+    dataFields: {
+      catnewsletter,
+      dognewsletter,
+    },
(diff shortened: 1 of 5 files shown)


blocks/newsletter-signup/newsletter-signup.js
```diff
@@ -0,0 +1,133 @@
...
+      },
+      mergeNestedObjects: true,
+      createNewFields: true
+    };
+
+    const fetchOpts = {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+        'Api-Key': '3e7a9624572b4827b156af44e72fceaa'
```
mfrisbey, line 104: Should we be storing the api key and/or API URL as a setting somewhere?

blocks/newsletter-signup/newsletter-signup.js
```diff
@@ -0,0 +1,133 @@
...
+    const dognewsletter = document.getElementById(dogInputId).checked;
+    const catnewsletter = document.getElementById(catInputId).checked;
+    const email = String(document.getElementById(emailInputId).value).trim();
+    const name = String(document.getElementById(nameInputId).value).trim();
+
+    showElement(inputError, false);
+    showElement(apiError, false);
+    showElement(success, false);
+
+    if (!email || !name) {
```
mfrisbey, line 84: Right now we're only checking that they provided an email. Are there any other projects where we've validated that they've entered a valid email format?

blocks/newsletter-signup/newsletter-signup.js
```diff
@@ -0,0 +1,133 @@
...
+      },
+      mergeNestedObjects: true,
+      createNewFields: true,
+    };
+
+    const fetchOpts = {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+        'Api-Key': '<API KEY HERE>',
```
mfrisbey, line 104: Should we be storing the api key and/or API URL as a setting somewhere?

scripts/scripts.js
```diff
@@ -664,6 +665,16 @@ async function optimizedBatchLoading(promises) {
   return Promise.all(promises.map((promise) => promise()));
 }
 
+async function loadNewsletter(footer) {
+  const title = document.createElement('h2');
+  title.innerText = 'Get the best of PetPlace straight to your inbox.';
```
ramboz, line 670: Any chance we can move that into the document? I'm not a huge fan of hard-coded strings. It makes it harder for authors to change them

blocks/newsletter-signup/newsletter-signup.js
```diff
@@ -0,0 +1,133 @@
...
+  buttonContainer.append(button);
+  target.append(buttonContainer);
+
+  const terms = document.createElement('span');
+  terms.classList.add('newsletter-terms');
+  terms.innerHTML = `
+    <p class="newsletter-success hidden">Thank you for signing up! We've received your information and will start sending newsletters to you.</p>
+    <p class="newsletter-error newsletter-input-error hidden">Please enter your name, provide a valid email address, and select at least one newsletter.</p>
+    <p class="newsletter-error newsletter-api-error hidden">Unfortunately, there was an error submitting your information. Please try again shortly.</p>
+    <p>By signing up, you agree to our <a href="/terms-of-use" target="_blank">Terms of Use</a> and <a href="/privacy-policy" target="_blank">Privacy Policy</a>.</p>
```
ramboz, line 130: I'd move this one into the block as well, and probably show on all versions (even if the mobile mocks don't have it), since it's a legal disclaimer.

blocks/newsletter-signup/newsletter-signup.js
ramboz: No, I think you can safely inline it here. it's going to end up in the code on the browser side no matter what

blocks/newsletter-signup/newsletter-signup.js
```diff
@@ -0,0 +1,133 @@
...
+    target.classList.add('hidden');
+  }
+}
+
+export default function decorate(block) {
+  const target = block.children[0];
+  const dogInputId = addCheckbox(target, 'dog', 'Yes, Send Me The Paw Print Newsletter.');
+  const catInputId = addCheckbox(target, 'cat', "Yes, Send Me The Cat's Meow Newsletter.");
+  const emailInputId = addTextbox(target, 'email', 'Type your email');
+  const nameInputId = addTextbox(target, 'name', 'Enter your name');
```
ramboz, line 57: To build the form, I'd suggest trying to re-use the regular form block with definition in an excel sheet.
https://www.hlx.live/developer/forms

You can probably export the `createForm` method, and move all the field definitions to the excel sheet.
We'd just need to pass an optional `onSubmit` handler that overrides the default one so you can have your custom logic.

What do you think?

It's a tad more involved, but ultimately gives

blocks/newsletter-signup/newsletter-signup.js
```diff
@@ -0,0 +1,133 @@
...
+  button.title = 'Sign up now';
+  button.classList.add('button', 'primary');
+  button.innerText = 'Sign up now';
+  button.addEventListener('click', async (e) => {
+    e.preventDefault();
+    enableElement(button, false);
+
+    const success = document.querySelector('.newsletter-signup .newsletter-success');
+    const inputError = document.querySelector('.newsletter-signup .newsletter-input-error');
+    const apiError = document.querySelector('.newsletter-signup .newsletter-api-error');
```
ramboz, line 73: These messages could potentially also be moved into the excel sheet

mfrisbey:
@ramboz I've applied all the feedback and converted this to a PR. The only feedback I haven't addressed is your [comment](https://github.com/hlxsites/petplace/pull/297#discussion_r1291432471) about moving some additional text to the block. The text in question is for the newsletter block that is auto-blocked in the footer - so we're building the block itself there.

I agree that it would be nice to let the users control this, but it would end up being a block within a block (i.e. a Newsletter Signup block inside the footer block). Do you have recommendations for how we could do this? Would we make a content fragment with the newsletter signup block definition?

ramboz:
Nice! 👏🏼 Yes, I think a fragment in that case would be the easiest. It's also something we can handle separately in a follow-up PR if you prefer to close the scope here.

blocks/newsletter-signup/newsletter-signup.js
```diff
@@ -0,0 +1,67 @@
...
+      showMessage(block, fd.Success);
+    }
+  } catch (e) {
+    showError(block, fd);
+  }
+}
+
+export default async function decorate(block) {
+  const form = await createForm(
+    `${window.location.protocol}//${window.location.host}/newsletter.json`,
```
ramboz, line 50: Would `/newsletter.json` be enough?

ramboz (approved):
Looks good to me!

Some thoughts:
1. we may need to make the form decoration "optional" based on page metadata. https://issue-257--petplace--hlxsites.hlx.page/newsletter-signup for instance probably shouldn't have it and instead show the form inline
2. once submitted, we probably should set a cookie and then not show the form once again for users that already registered

blocks/newsletter-signup/newsletter-signup.js
mfrisbey: I tried that, but the `URL()` constructor used in `createForm()` doesn't like it. I could refactor a couple things to get rid of it pretty easily, though.

ramboz: No, all fine then. was just curious :)

mfrisbey:
> @ramboz we probably should set a cookie

Do you have any code examples where a Franklin site is working with cookies?

ramboz:
Not readily available. also, it could be local storage instead of cookies. might actually be better as we don't send it back to the server so it would be cleaner for privacy reasons

mfrisbey:
@ramboz added a few extra things:

* Shortened the form URL to `/newsletter.json`
* Added a metadata property to prevent the newsletter signup from showing in the footer.
* Set a value in local storage after the user signs up, and don't display the newsletter signup in the footer after that.

PR #302, opened 2023-08-18
feat: replace basic search with a full-text search engine

ramboz:
This PR introduces `elasticlunr` to perform full-text search (think of `elasticsearch` but client-side). The search is performed in a worker so that loading the index and filtering results doesn't impact the main thread.

To speed things up, a search-index is pre-build whenever a new article is published, and pushed to the git repo.
And for slower connections (2G), we fall back to a lighter simple plain-text search on the last 5K articles to speed things up as loading the whole search index is quite slow.

Test URLs:
- After:
  - https://elasticlunr--petplace--hlxsites.hlx.page/
  - https://elasticlunr--petplace--hlxsites.hlx.page/search?martech=off
  - https://elasticlunr--petplace--hlxsites.hlx.page/search?query=dog+ticks&martech=off


FentPams:
Looks awesome to me!!!! Good job!👍

.github/ci/build-search-index.js
```diff
@@ -0,0 +1,23 @@
...
+  this.addField('category name');
+  this.addField('author');
+
+  json.data.forEach((doc) => this.addDoc(doc));
+});
+
+// Save the search index to a static file in the repository
+const targetDirectory = process.argv[2];
+const targetFile = `${targetDirectory}/search-index.json`;
+fs.writeFileSync(targetFile, JSON.stringify(idx));
```
mfrisbey, line 23: Where does this end up being saved? I looked in their sharepoint because I was curious how big the file is, but I don't see it there.

.github/workflows/pre-build-search-index.yaml
```diff
@@ -0,0 +1,50 @@
+name: Pre-build the seach index
+
+on:
+  repository_dispatch:
+    types:
+      - resource-published
```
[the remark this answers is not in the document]
ramboz: Yes, whenever the pipeline publishes a new document, it will trigger this event on the repo

.github/ci/build-search-index.js
ramboz: We can't directly write to sharepoint unfortunately, as this requires proper oauth and 2fa typically, so what we do instead is that we push the file to the git repo which only requires a personal access token.

The file is 20MB uncompressed but ends up being sitting at ~3.3MB once gzipped on the page.
On desktop, that gives us ~100ms delay, on 4G 1.5~2s and on 3G we are up to 20secs.

ramboz: Did a few more optimizations to minimize the index. Now we are down to 14MB and ~2.7MB gzipped:

- ~70ms on desktop
- ~1.4s on 4G
- ~17s on 3G

PR #305, opened 2023-08-28
feat: add conversion tracking

ramboz:
This PR adds the conversion plugin and tracks key conversions:
- ad clicks
- Form submissions
- pet insurance quote clicks

Test URLs:
- After:
  - https://track-conversion--petplace--hlxsites.hlx.live/
  - https://track-conversion--petplace--hlxsites.hlx.live/?martech=off


blocks/ad/ad.js
```diff
@@ -80,6 +81,25 @@ const loadedObserver = new MutationObserver((entries) => {
   });
 });
 
+let conversionMonitor;
```
ramboz, line 84: Since ads are instrumented via iframes, we cannot directly listen for a `click` event since those wouldn't bubble up to the parent window. Instead, we use a polling mechanism that checks the currently focused/active element in the DOM. If this changes to an iframe in an `ad` block, we can safely assume a click happened

scripts/scripts.js
```diff
@@ -705,6 +705,12 @@ export function setNewsletterSignedUp() {
  */
 async function loadLazy(doc) {
   const main = doc.querySelector('main');
+
+  const context = { getMetadata, toClassName };
```
ramboz, line 709: We instrument this early before `loadBlocks`, so that the `convert` method exists by the time we render the blocks. That way instrumentation is easier

mfrisbey (approved):
Nice! It's interesting to see how that works

PR #306, opened 2023-08-28
feat: add popup for newsletter signup

### LINKED ISSUE #257: feat: newsletter subscription

mfrisbey:
Adds a popup to the footer that will prompt visitors to sign up for the newsletter.

* Adds a new popup block that will display its children in a modal dialog.
* The popup block utilizes aria-dialog for accessibility.
* Auto block a popup containing the newsletter signup block in the site's footer.
* Normally, the aria-dialog functionality will add a button that, when clicked, displays the dialog. This button is still present on the site, but has been style to a `fixed` position at the top of the page, but with a `left` value that puts it out of the viewport. This makes it so that when the dialog is closed, the page scroll position will remain in the same place.
* The dialog itself is also at a `fixed` position. This is so that the dialog will appear wherever the visitor is currently looking on the page.
* Added a new `/configuration.xlsx` file with a sheet for the newsletter popup. This file defines how long the popup will wait before displaying, and pages where the popup should not be included.

When rendered, the popup looks like this:

Screenshot 2023-08-28 at 2 55 00 PM

Fix #257

Test URLs:
- After:
  - https://issue-257-popup--petplace--hlxsites.hlx.page/
  - https://issue-257-popup--petplace--hlxsites.hlx.page/?martech=off


blocks/popup/popup.css
```diff
@@ -0,0 +1,41 @@
+.popup div[role="dialog"][aria-hidden="false"] {
+  position: fixed;
+  top: 0;
+  bottom: 0;
+  left: 0;
+  right: 0;
```
ramboz, line 6: You could simplify with:
```suggestion
  inset: 0
```

blocks/popup/popup.js
```diff
@@ -0,0 +1,23 @@
...
+  const target = block.children[0].children[0];
+  [...target.children].forEach((child) => {
+    contentContainer.append(child);
+  });
+  ariaDialog.append(contentContainer);
+  target.append(ariaDialog);
+
+  setTimeout(() => {
+    showDialog.click();
+  }, 1000);
```
ramboz, line 22: You could probably set a `0` duration or use `window.requestAnimationFrame`

styles/styles.css
```diff
@@ -784,6 +784,20 @@ a.sr-focusable:focus {
   font-size: var(--body-font-size-xxs);
 }
 
+/* stylelint-disable-next-line no-descending-specificity */
+footer .popup hlx-aria-dialog > button {
```
ramboz, line 788: This made me realize that the use case is to have a dialog with no button that triggers it.
This is actually closer to the https://www.w3.org/WAI/ARIA/apg/patterns/alertdialog/ pattern than the regular https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/

I would go ahead here, but file a separate issue in the backlog to convert that to the proper pattern down the road so we don't need a fictive trigger button, and also have something which is a bit more accessbile

ramboz (approved):

blocks/popup/popup.js
mfrisbey: I think this was leftover debug code, I've removed the `setTimeout()` call.

styles/styles.css
mfrisbey: Created #308

scripts/glob-regex/glob-regex.js
```diff
@@ -0,0 +1,34 @@
...
+function join(globs) {
+  return `((${globs.map(replace).join(')|(')}))`;
+}
+
+function globRegex(glob) {
+  return new RegExp(`^${(Array.isArray(glob) ? join : replace)(glob)}$`);
+}
+
+globRegex.replace = replace;
+export default globRegex;
```
mfrisbey, line 34: I added this file here because the module itself was short, and I wasn't sure how to add a node dependency. Please let me know if there's a better way to do this.

scripts/scripts.js
```diff
@@ -699,6 +726,52 @@ export function setNewsletterSignedUp() {
...
+ * @param {string} sheet The name of the sheet whose configuration should
+ *  be loaded.
+ * @returns {Promise<Array<*>|undefined>} Resolves with the array of
+ *  configuration values from the specified sheet. Will be undefined
+ *  if the configuration could not be found.
+ */
+export async function getConfiguration(sheet) {
+  let config;
+  try {
+    config = await ffetch('/drafts/frisbey/configuration.json')
```
mfrisbey, line 740: Please take a look at this configuration file and let me know if you have feedback. If not, I'll move it to the root and preview it.

blocks/popup/popup.css
```diff
@@ -0,0 +1,38 @@
...
+  position: fixed;
+  inset: 0;
+  margin: 0;
+  padding: 0;
+  background-color: var(--color-purple-transparent);
+  z-index: 1000;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+}
```
ramboz, line 11: In theory, we only need te `display` prop to have higher specificity, so we could also do:
```suggestion
.popup div[role="dialog"] {
  position: fixed;
  inset: 0;
  margin: 0;
  padding: 0;
  background-color: var(--color-purple-transparent);
  z-index: 1000;
  justify-content: center;
  align-items: center;
}
.popup div[role="dialog"][aria-hidden="false"] {
  display: flex;
}
```

scripts/glob-regex/glob-regex.js
ramboz: No, that is perfect 👍🏼 
I would maby just add a header comment that points back to the source (i.e. the github repo) so that it's easier to bring back updates in this if ever needed

scripts/scripts.js
ramboz: I'd also switch to plain fetch as it is unlikely we will ever need pagination on this.
So just:
```js
  const response = await fetch(`/drafts/frisbey/configuration.json?sheet=${sheet}`);
  const json = await response.json();
  config = json.data;
```

You might have to explicitly make it a multi-sheet workbook by adding another dummy sheet for now (like `helix-general` with just the header row)

scripts/scripts.js
```diff
@@ -699,6 +726,52 @@ export function setNewsletterSignedUp() {
...
+}
+
+async function addNewsletterPopup() {
+  const newsletterConfig = await getConfiguration('newsletter-popup');
+  if (!newsletterConfig) {
+    return;
+  }
+  const delay = newsletterConfig
+    .find((item) => item.Key === 'delay-seconds');
+  const seconds = delay ? parseInt(delay.Value, 10) : 10;
```
ramboz, line 757: You can likely just use `Number(delay.Value)`, `parseInt` is typically used when you have numbers+units you want to parse (i.e. `10px`, `5lbs`, etc.)

ramboz (approved):
Looks good to me! Left a few suggestions

PR #307, opened 2023-08-29
feat: add RSS/Atom feed generation on article publishing

ramboz:
Leverages the `feed` node module to build RSS & atom XML feeds for all articles. Those files are updated along with the search index when new articles are published.

### DIFF
--- a/.github/workflows/pre-build-search-index.yaml
+++ b/.github/workflows/pre-build-search-index.yaml
@@ -1,4 +1,4 @@
-name: Pre-build the seach index
+name: Pre-build the seach index and update RSS/Atom feeds
 
 on:
   repository_dispatch:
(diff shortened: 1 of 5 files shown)


.github/workflows/pre-build-search-index.yaml
```diff
@@ -43,8 +43,12 @@ jobs:
       working-directory: ./.github/ci
       run:  node build-search-index.js ${{ env.TARGETDIRECTORY }}
 
+    - name: Generate the search index
```
mfrisbey, line 46: Should this be labeled differently?

mfrisbey (approved):

.github/workflows/pre-build-search-index.yaml
ramboz: Good catch 👍🏼

PR #315, opened 2023-09-07
feat: create subscription confirmation page

### LINKED ISSUE #314: feat: implement https://www.petplace.com/subscription-confirmation/
Image

Image

Image

mfrisbey:
Created a variant of the columns block called "emphasized":

Screenshot 2023-09-11 at 3 49 24 PM

Rendered as:

Screenshot 2023-09-11 at 3 49 45 PM

The original page had some element placements that would have required a new template or section style, so I took some liberties and used a blade instead. Due to that, the only new code required for the page was the new variant. Let me know your thoughts on the layout, and I'll also run it by the petplace team.

This will require publishing the redirects spreadsheet and the new subscription-confirmation word doc. There was a self-referencing redirect, so the URL was stuck in a redirect loop; I removed the unneeded references.

Fix #314 

Test URLs:
- Before: https://main--petplace--hlxsites.hlx.page/subscription-confirmation
- After:
  - https://issue-314--petplace--hlxsites.hlx.page/subscription-confirmation
  - https://issue-314--petplace--hlxsites.hlx.page/subscription-confirmation?martech=off


ramboz (commented):
I wonder if we could easily re-use the columns block for this maybe, and possibly just use a variant of the block for the styling, like `Columns (emphasized)` or similar

mfrisbey:
@ramboz I think we could use the existing column layout. I'll look into it.

ramboz (approved):

PR #318, opened 2023-09-11
feat: capture newsletter errors as rum data.

mfrisbey:
As requested by petplace team, capture details about errors that occur when attempting to submit newsletter data to iterable.

Test URLs:
- After:
  - https://rum-error--petplace--hlxsites.hlx.page/
  - https://rum-error--petplace--hlxsites.hlx.page/?martech=off

### DIFF
--- a/scripts/scripts.js
+++ b/scripts/scripts.js
@@ -981,6 +981,14 @@ export async function createBreadCrumbs(crumbData, chevronAll = false) {
   return breadcrumbContainer;
 }
 
+/**
+ * Logs the information about an error encountered by the site.
+ * @param {Error} e Error information to log.
+ */
+export async function captureError(e) {
+  sampleRUM('error', e);
+}
+
 async function loadPage() {
   await loadEager(document);
   await loadLazy(document);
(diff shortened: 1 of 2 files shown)


scripts/scripts.js
```diff
@@ -981,6 +981,14 @@ export async function createBreadCrumbs(crumbData, chevronAll = false) {
   return breadcrumbContainer;
 }
 
+/**
+ * Logs the information about an error encountered by the site.
+ * @param {Error} e Error information to log.
+ */
+export async function captureError(e) {
+  sampleRUM('error', e);
```
ramboz, line 989: RUM works by specifying a `source` and `target` property in the second parameter. Typically, the `source` is the identifier for the data point and `target` is the value.
I'd probably go with:
```suggestion
  sampleRUM('error', { source: 'newletter-signup', target: e.message);
```

mfrisbey:
@ramboz I modified the rum payload (thanks for the tip), and used `?rum=on` to verify that the request is being sent. Does everything look ok to you?

ramboz (approved):
Yes, all good!

PR #329, opened 2023-10-05
feat: added pushly service worker

ankyc-phi:
Commits :
- got service worker file from cdn
- added sw script to pushly code
- added sw file to lint ignore
 
Jira Link : https://pethealthinc.atlassian.net/browse/PPL-26

Test URLs:
- After:
  - https://ppl-26-pushly-sw--petplace--hlxsites.hlx.page/
  - https://ppl-26-pushly-sw--petplace--hlxsites.hlx.page/?martech=off

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -4,3 +4,4 @@ ffetch.js
 lite-yt-embed
 lite-tiktok
 partytown
+pushly-sdk-worker.js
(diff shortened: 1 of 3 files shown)


ramboz (commented):
The code looks good to me, but I don't see the service worker being activated for some reason. We may need to debug this further and check the documentation again if we are maybe missing something.

ramboz:
I managed to track it down to this error:
```
ExitException: Current host is not whitelisted for use
```

PR #331, opened 2023-10-13
pushing changes for review

ankyc-phi:
draft PR to get visibility on datalayer approach

Test URLs:
- After:
  - https://datalayer-ppl-17-25--petplace--hlxsites.hlx.page/
  - https://datalayer-ppl-17-25--petplace--hlxsites.hlx.page/?martech=off

### DIFF
--- a/scripts/datalayer.js
+++ b/scripts/datalayer.js
@@ -1,0 +1,57 @@
+window.dataLayer = window.dataLayer || [];
+
+// document.onreadystatechange = () => {
+//   if (document.readyState === 'complete') {
+//     console.log('ready')
+//     handleGlobalVariables()
+//     handleSearch()
+//   }
+// };
+
+// GLOBAL VARIABLES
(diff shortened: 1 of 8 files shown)


scripts/datalayer.js
ramboz, line 1: I'd be tempted to move this to the individual blocks actually, in `/blocks/header`, `/blocks/search` and `blocks/newsletter-signup` so it sits with the actual implementation. We save ourselves an additional resource to download that way, and it stays in the context.

@meryllblanchet @mfrisbey what are your thoughts?

ankyc-phi: this can work for the ones that we for sure know are in blocks and are a one-off case, i wonder what it would look like for implementation when it comes to article share or element clicks

ankyc-phi:
closed this one due to branch renaming, new one is here - https://github.com/hlxsites/petplace/pull/332

PR #333, opened 2023-10-25
Datalayer implementation for tracking article shares

ankyc-phi:
Branch Commits :

- changes to href for consistency
- added eslint rule for ++ in for loops
- small changes to datalayer file
- handle article share push to DL
- added try catch block for console error
- change to foreach instead of for loop

Jira Issues :
https://pethealthinc.atlassian.net/jira/software/c/projects/PPL/boards/35?selectedIssue=PPL-19

Test URLs:
- After:
  - https://datalayer-ppl-19--petplace--hlxsites.hlx.page/
  - https://datalayer-ppl-19--petplace--hlxsites.hlx.page/?martech=off

### DIFF
--- a/scripts/datalayer.js
+++ b/scripts/datalayer.js
@@ -14,8 +14,32 @@ const handleGlobalVariables = () => {
   });
 };
 
-// ARTICLE SHARE -> to come in later
+// ARTICLE SHARE
+const handleArticleShare = () => {
+    const aTags = document
+      .querySelector('.social-share')
+      .getElementsByTagName('a');
+
+    for (let i = 0; i < aTags.length; i++) {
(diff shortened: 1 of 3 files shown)


scripts/datalayer.js
```diff
@@ -1,6 +1,5 @@
-window.dataLayer ||= [];
-
 export const pushToDataLayer = (layer) => {
+  window.dataLayer ||= [];
```
ramboz, line 2: I would keep this outside the function so it's executed once when the file is loaded, rather than every time you push to the datalayer.

ramboz (approved):
Overall good, but left a comment

ramboz (approved):

PR #339, opened 2023-11-06
Datalayer implementation for CTA buttons and fix for nav bar

ankyc-phi:
Branch Commits :

1. New Commits
- added in tracking for cta buttons
- checking for button on home page only

2. Fixes/Adjustments
- tracking header clicks in menu bar
- created some helpers for header

Jira Issues :
https://pethealthinc.atlassian.net/jira/software/c/projects/PPL/boards/35?selectedIssue=PPL-44
https://pethealthinc.atlassian.net/jira/software/c/projects/PPL/boards/35?selectedIssue=PPL-43
https://pethealthinc.atlassian.net/jira/software/c/projects/PPL/boards/35?selectedIssue=PPL-18

Test URLs:
- After:
  - https://datalayer-ppl-44-43--petplace--hlxsites.hlx.page/
  - https://datalayer-ppl-44-43--petplace--hlxsites.hlx.page/?martech=off

### DIFF
--- a/scripts/datalayer.js
+++ b/scripts/datalayer.js
@@ -39,8 +39,14 @@ const handleElementClicks = () => {
   headerNavHelper();
   headerMenuHelper();
   headerSidebarHelper();
+    const btnTracking = document.querySelectorAll(
+      '.slides-container .button-container a'
+    );
+    btnTracking.forEach((btn) => {
+      btn.addEventListener('click', () => {
+        clickHelper('CTA Button', btn.title, 'button', btn.href);
+      });
     });
-  });
 
   // footer links
   const observer = new MutationObserver((entries) => {
--- a/scripts/datalayer.js
+++ b/scripts/datalayer.js
@@ -39,6 +39,9 @@ const handleElementClicks = () => {
   headerNavHelper();
   headerMenuHelper();
   headerSidebarHelper();
+
+  // cta buttons (homepage only)
+  if (window.location.pathname === '/') {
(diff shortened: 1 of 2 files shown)


ankyc-phi:
> Seeing you are adding quite a bunch of listeners in all those PRs or the datalayer, I'd be tempted to switch to the previous proposal we had to have a single global click listener and then filter on the target that emitted the event.
> 
> You can then leverage `ev.target.closest(…)` to check in which block you are and then adjust the parameters for the `clickHelper` accordingly

thanks for the comment, that's definitely an approach we should take - do you think i should scrap this PR and work on those changes right away, or do the one listener after this goes in so that i can set up the global listener before working on the article links??

ramboz:
It won't make much of a difference. It's really up to you.

Personally, I would leverage this PR to do the refactoring and adding the new cases you want handled in it as well. Each separate listener you add does take a (minor) toll on the CPU, so ultimately those add up and degrade the responsiveness of the page.

ankyc-phi:
> It won't make much of a difference. It's really up to you.
> 
> Personally, I would leverage this PR to do the refactoring and adding the new cases you want handled in it as well. Each separate listener you add does take a (minor) toll on the CPU, so ultimately those add up and degrade the responsiveness of the page.

going to actually close this PR and open up a new one as the header stuff is fixed in the one i already have up, i'll create a different branch for the cta buttons and a small change for article sharing

https://github.com/hlxsites/petplace/pull/342

PR #347, opened 2023-11-17
added margin to create some space on paragraphs

ankyc-phi:
Jira Issue: https://pethealthinc.atlassian.net/jira/software/c/projects/PPL/boards/35?selectedIssue=PPL-53

Test URLs:
- Before: https://main--petplace--hlxsites.hlx.page/newsletter-signup
- After:
  - https://newsletter-page-design-changes--petplace--hlxsites.hlx.live/newsletter-signup
  - https://newsletter-page-design-changes--petplace--hlxsites.hlx.live/newsletter-signup?martech=off

Screenshot:
image

### DIFF
--- a/templates/article-signup/article-signup.css
+++ b/templates/article-signup/article-signup.css
@@ -11,8 +11,8 @@
   margin-top: 1rem;
 }
 
-.article-signup-content p {
-  margin: 20px auto;
+.article-signup main .article-signup-content p {
+  margin-bottom: 1rem;
 }
 
 @media (min-width: 1000px) {
--- a/templates/article-signup/article-signup.css
+++ b/templates/article-signup/article-signup.css
@@ -11,6 +11,10 @@
   margin-top: 1rem;
 }
 
+.article-signup-content p {
+  margin: 20px auto;
+}
+
 @media (min-width: 1000px) {
   .article-signup main .article-signup-section {
     grid-template:
--- a/templates/article-signup/article-signup.css
+++ b/templates/article-signup/article-signup.css
@@ -11,7 +11,7 @@
   margin-top: 1rem;
 }
 
-.article-signup main .article-signup-content p {
+.article-signup .article-signup-content p {
   margin-bottom: 1rem;
 }
 


ramboz (approved):
Changes look good to me!

On a side note:
- we typically use `rem` instead of `px` in the project, as this makes zooming in cleaner on older browsers (for accessibility reasons). We used `margin-bottom: .75rem` on other templates for this
- we typically narrow the selector to `<template> .default-content-wrapper p` so it doesn't bleed into other blocks

ankyc-phi:
> Changes look good to me!
> 
> On a side note:
> 
> * we typically use `rem` instead of `px` in the project, as this makes zooming in cleaner on older browsers (for accessibility reasons). We used `margin-bottom: .75rem` on other templates for this
> * we typically narrow the selector to `<template> .default-content-wrapper p` so it doesn't bleed into other blocks

thanks for calling this out, will make the change and push an update to this before EOD

templates/article-signup/article-signup.css
```diff
@@ -11,6 +11,10 @@
   margin-top: 1rem;
 }
 
+.article-signup main .article-signup-content p {
```
ramboz, line 14: I'd remove the `main` selector as it won't help much here. There will always be a `main` in the path, and there will always only be one on the page. As it is, it will just slightly slow done the selector since the browser has to essentially parse all tags on the page to check if we have a match, and we know it's going to match 100% of the time.

ramboz (approved):
Looks good.

Though we can fine-tune even further

templates/article-signup/article-signup.css
ankyc-phi: okay sure, makes sense - i was just following the format as the other ones in the same file, do you want me to make changes to the rest of the css file as well??

ramboz: Actually, yes, you can :) but again, it's not a blocking concern, just a suggestion

ankyc-phi: sounds good, i've made a ticket for that on our end to review the css files :)

ankyc-phi: so this PR will only have the small change with the <main> selector removed

PR #350, opened 2023-11-30
Styling for Insurance Landing Page

Tom-Scott-Hero:
Branch Commits:

- Pushing Initial Insurance Landing Page Styles
- Mobile Fixes
- Adding more mobile/tablet styles
- Updating Responsive Styles
- datalayer tracking in lazy phase
- adding some changes to center content
- removing main selector as it's no longer needed
- Responsive Updates to insurance landing page

Jira Issue:
https://pethealthinc.atlassian.net/browse/PPL-31

Test URLs:

- Before: 
     https://main--petplace--hlxsites.hlx.page/pet-insurance
- After:
     https://insurance-landing--petplace--hlxsites.hlx.page/pet-insurance

### DIFF
--- a/blocks/details-box/details-box.css
+++ b/blocks/details-box/details-box.css
@@ -1,0 +1,120 @@
+
+/* Details Box Section Block */
+.details-box.rounded {
+  padding: 2.5rem;
+  margin: 0 auto 56px !important;
+  border-radius: 20px;
+  background-color: var(--color-dark-maroon); 
+  @media (min-width: 800px){
+    padding: 2rem;
+  }
+  @media (min-width: 1024px){
+    padding: 2.5rem;
+  }
+}
+
+/* Details Box  Title */
+.details-box h4 {
+  border-bottom: 2px dotted #fff;
+  color: var( --text-color-inverted);
+  text-align: center;
+  font-family: var(--body-font-family);
+  font-style: normal;
+  font-weight: 700;
+  margin-bottom: 0;
+  padding: 0;
+  padding-bottom: 2rem;
+  
+  font-size: var( --heading-font-size-xs);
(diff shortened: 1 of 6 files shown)


mfrisbey (commented):
I'll approve it if you'd like, but there seems to be some incorrect block implementation, which is causing a lot of 404's in the console. For example, there are `subhead`, `reasons`, and `insurance` blocks in the word document, but there is no corresponding code for those. Normally, to implement these we would create this file structure:

```
blocks
..subhead
....subhead.css
....subhead.js
..reasons
....reasons.css
....reasons.js
..insurance
....insurance.css
....insurance.js
```

Then we would add CSS specific to a block in its corresponding `*.css` file. It looks like we've got all the block styles in `insurance-landing-page.css`, whereas normally they would be separated into the block files.

The way it is _works_, but we've got 404's in the console, and blocks that couldn't be reused on other pages because all the block styles are in the template instead of the block files.

ramboz (approved):
I do agree with @mfrisbey recommendations on re-using variables and px vs rem, but this isn't a blocker and ultimately down to your preferences maintaining the codebase long-term… the block implementations and 404 are more a concern since they will negatively impact performance and possibly ranking because we have errors in the browser logs essentially

mfrisbey:
Thanks @Tom-Scott-Hero - looks great! The last thing to check is that the lint workflow is failing. You can see the errors locally by running the command `npm run lint` in the root directory. After you've fixed those and the workflows are green, we'll be ready to merge!

Tom-Scott-Hero:
@mfrisbey Thanks! I just pushed an update with the linter corrections. Let me know if there are any other changes needed.

mfrisbey (approved):
Thanks!

PR #360, opened 2024-01-19
feat: ads for middle of article sections

ankyc-phi:
Branch Commits:
- moved adsense out of helpers
- moved helper functions into utils file
- removed advertising text in the block
- display ad for article middle section

Jira Issue:
https://pethealthinc.atlassian.net/browse/PM-87

Test URLs:
- After:
  - https://feat-pm-87-article-middle-ads--petplace--hlxsites.hlx.live/
  - https://feat-pm-87-article-middle-ads--petplace--hlxsites.hlx.live/?martech=off

### DIFF
--- a/blocks/search/search.css
+++ b/blocks/search/search.css
@@ -7,6 +7,55 @@
     margin: 0 auto;
 }
 
+.section.search-container h1 {
+    display: inline-block;
+    top: -60px;
+    background-color: var(--text-color-inverted);
+    border-radius: 40px;
+    padding: 40px 30px 0;
+    margin: 0 ;
+    bottom: 60px;
+    width: 60%;
+    color: black;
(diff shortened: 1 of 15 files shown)


ramboz (commented):
Looks good overall.

A few suggestions:
1. we initially had ads instrumented via `/blocks/ad/ad.js` and definitions done in `/ads.xslx` in sharepoint. we now have 2 different ways to handle ads, so I'd consider refactoring at some point to have a single way of doing it
2. the ads instrumentation we did also captures clicks in our own RUM data. See: https://github.com/hlxsites/petplace/blob/main/blocks/ad/ad.js#L83-L100. by using a different instrumentation mechanism, we do lose some of that data collection

ramboz (approved):

PR #362, opened 2024-01-19
PM-99: Adding ability to click the Search Icon in the Search Bar

Tom-Scott-Hero:
Adding the ability to click on the Search Icon inside the search bar in both the header Header and Hamburger search bars. 

Fix #PM-99

Test URLs:
- Before: https://main--petplace--hlxsites.hlx.page/article/category/pet-behavior-training?martech=off
- After:
  - https://pm-99--petplace--hlxsites.hlx.live/article/category/pet-behavior-training?martech=off
  - https://pm-99--petplace--hlxsites.hlx.live/article/category/pet-behavior-training


ramboz (commented):
While this technically works, it won't be considered properly accessible in the WCAG 2.0 sense.

Ideally, we should split this into a proper `input[type="search"]` + `button[type="submit"]`, where the button can be focused (via tab key) and would also take the click event. Essentially something as described in [type="search"], but just styled as a single widget.
We had an old PR that never got done https://github.com/hlxsites/petplace/pull/299/files which was on its way there if you want to take some inspiration.

At the moment, adding the click event to the form only works if the user doesn't actually hit the form element by mistake. For instance, with the rounder border, if you click just in the angles, you'd actually trigger the search. I'd also safeguard by checking that we actually have a value in the search field before submitting the form

Tom-Scott-Hero:
Hey @ramboz, great suggestion. I updated the code to use a Button element based on the previous PR you mentioned. I also added checks to the submission that check for a empty search string before submitting the form.

ramboz:
@Tom-Scott-Hero Great! It seems you still have some minor linting issues in the code:
https://github.com/hlxsites/petplace/actions/runs/7628715650/job/20780433335?pr=362#step:5:25

Looks great otherwise 👍🏼

Tom-Scott-Hero:
Hey @ramboz, Thanks, i didn't notice the linter error. I added the new line to lazy-styles.css and hopefully this will resolve things. Thank you

ramboz:
@Tom-Scott-Hero You can run `npm install` at the root of your repo, which should install ESLint and Stylelint in the project. Depending on your IDE this might be enough for it to directly show you the errors while you edit the files (VS Studio Code does) or you might have to install an additional extension for these to show up.
Any any case, you can also run `npm run lint` at the root of the project to get the same run as the CI pipeline does directly in the CLI.

ramboz (approved):

PR #365, opened 2024-01-23
refactor: use placeholders for static strings to prepare for multi-language support

ramboz:
As part of #364, we need to extract all hardcoded strings so they can be localized as needed. We also remove all the testing logic at the same time since this isn't used anymore so we have a cleaner baseline for future regional sites.

For this, a new `/placeholders.xslx` sheet has been added to Sharepoint, and the site now pulls the static strings from there. We'll be able to transparently replace the `placeholders.xslx` sheet for each locale down the road so we have properly localized strings across the website.

We also introduce a new `getPlaceholder(key, options)` method in the `scripts.js` so we can easily pull strings and quickly inject variables in them (uses a `{{variableName}}` syntax)

Test URLs:
- After:
  - https://placeholders--petplace--hlxsites.hlx.live/
  - https://placeholders--petplace--hlxsites.hlx.live/?martech=off


blocks/header/header.js
```diff
@@ -139,27 +144,29 @@ export default async function decorate(block) {
...
+    toggle.setAttribute('aria-label', isExpanded
+      ? getPlaceholder('closeNavItem')
+      : getPlaceholder('openNavItem', { item: target.textContent }));
   });
   navSidebar.querySelectorAll('[role="tree"] [role="treeitem"]').forEach((item) => {
     observer.observe(item, { attributes: true });
   });
 
-  block.querySelectorAll('.nav-sidebar-social a').forEach((a) => {
+  block.querySelectorAll('.nav-sidebar-social a').forEach(async (a) => {
```
mfrisbey, line 166: There are a few places where I noticed the addition of `async`, but I don't think `getPlaceholder()` is an `async` method? Are those changes intentional?

mfrisbey (approved):
Looks good! I added a comment about some `async` additions that I noticed, but probably not anything that would cause problems.

blocks/header/header.js
ramboz: Good catch! Missed those when refactoring!

PR #395, opened 2024-02-19
feat: add region selector

ramboz:
Adds a region selector in the footer that will also automatically trigger if the detected region (guessed from the CDN pop location) does not match the region from the URL.

Test URLs:
- After:
  - https://language-selector--petplace--hlxsites.hlx.live/
  - https://language-selector--petplace--hlxsites.hlx.live/?martech=off
  - https://language-selector--petplace--hlxsites.hlx.live/en-gb/
  - https://language-selector--petplace--hlxsites.hlx.live/en-gb/?martech=off


mfrisbey (approved):
Nice!

PR #396, opened 2024-02-19
Feature/ppad 14 adoption landing dok

hero-dokane:
**Jira Ticket:** PPAD-14

**Purpose**: Adoption Landing Page functionality

**Demo**: http://localhost:3000/pet-adoption/


blocks/adopt-search/adopt-search.js
```diff
@@ -1,215 +1,339 @@
...
-                animalLocation.innerHTML = `${animal.City}`;
-                div.append(anchor);
-                div.append(animalName);
-                div.append(p);
-                div.append(animalLocation);
-                tempResultsBlock.append(div);
-            });
-        });
+async function callAnimalList(method, path, payload) {
+  const baseUrl = 'https://api-stg-petplace.azure-api.net/';
```
ericvangeem, line 68: @raymond-p-sanford @hero-dokane @jane-li-hero Can we abstract this API URL to some common module? Since these are Stage API hosts, and knowing at some point we'll need to switch to referencing Prod API hosts, it would be ideal if that determination is managed in one central place. We might have a common module that contains logic to return either a Stage API host or Prod API host based on the current site URL.

blocks/adopt-search/adopt-search.js
```diff
@@ -1,215 +1,339 @@
...
+  img.data = coverImagePath;
+  img.type = 'image/jpg';
+  img.className = 'pet-card-image';
+  const fallback = document.createElement('img');
+  fallback.src = getMetadata('image-fallback');
+  img.append(fallback);
+  const cardBody = document.createElement('div');
+  cardBody.className = 'pet-card-body';
+  cardBody.innerHTML = `
+      <h3 class='pet-card-name'><a href='/pet-adoption/pet/${clientId}/${animalId}' class='stretched-link'>${Name?.replace(
```
ericvangeem, line 42: @hero-dokane @raymond-p-sanford @jane-li-hero in the URL, replace `/pet` with the plural lower-cased animal type. Also, swap the `${animalId}` and `${clientId}` values in the URL. 
For example, for an animal with these values: 
- `Animal type` = "Dog"
- `animalId` = "abc"
- `clientId` = "123"  

then the URL should be `/pet-adoption/dogs/abc/123`

Since this pet card rendering is used in multiple places (landing page, search results, pet detail page), it might make sense to export this as a common module that other blocks can call.

ericvangeem (changes requested):

blocks/adopt-search/adopt-search.js
ericvangeem: Just sharing for reference, the Pet Detail URL is documented here https://herodigital.atlassian.net/wiki/spaces/PA/pages/3372122366/Pet+Details#URL-Structure

blocks/adopt-search/adopt-search.js
```diff
@@ -1,215 +1,350 @@
...
+export default async function decorate(block) {
+  const payload = {
+    locationInformation: {
+      clientId: null,
+      zipPostal: '92102',
+      milesRadius: 10,
+    },
+    animalFilters: {
+      startIndex: 0,
+      numResults: 4,
```
ericvangeem, line 325: @hero-dokane @raymond-p-sanford tech design is to request 50 `numResults` and use JS to randomly select 4 from that list. https://herodigital.atlassian.net/wiki/spaces/PA/pages/3371729078/Landing+Page#Adoptable-Pets

Example code snippet:
```
function getRandomObjects(array, count) {
  const shuffled = array.sort(() => 0.5 - Math.random());
  return shuffled.slice(0, count);
}

const objects = [
  { id: 1, name: 'Object 1' },
  { id: 2, name: 'Object 2' },
  { id: 3, name: 'Object 3' },
  { id: 4, name: 'Object 4' },
  { id: 5, name: 'Object 5' },
  { id: 6, name: 'Object 6' },
  { id: 7, name: 'Object 7' },
  { id: 8, name: 'Object 8' }
];

const randomObjects = getRandomObjects(objects, 4);
console.log(randomObjects);
```

ericvangeem (changes requested):

blocks/adopt-search/adopt-search.js
raymond-p-sanford: @hero-dokane This looks good now.  Thanks

blocks/adopt-search/adopt-search.js
raymond-p-sanford: Looks good

PR #402, opened 2024-02-22
feat: ads for tag page + refactoring adsense

ankyc-phi:
Branch Commits:
- creating tag divs and displaying ads
- refactored adsense functionality to be more streamlined
- using new adsense func for tag page
- using new adsense func for article page
- using new adsense func for home page
- using new adsense func for category page

Jira Issue:
https://pethealthinc.atlassian.net/browse/PM-89

Test URLs:
- After:
  - https://feat-tag-page-ads--petplace--hlxsites.hlx.live/
  - https://feat-tag-page-ads--petplace--hlxsites.hlx.live/?martech=off

Screenshots:

desktop
image

mobile
image

### DIFF
--- a/templates/tag-index/tag-index.js
+++ b/templates/tag-index/tag-index.js
@@ -10,6 +10,8 @@ import {
   isTablet,
   meterCalls,
 } from '../../scripts/scripts.js';
+import { pushToDataLayer } from '../../scripts/utils/helpers.js';
+import { adsDefineSlot, adsDivCreator } from '../../scripts/adsense.js';
 
 const PAGINATE_ON = 12;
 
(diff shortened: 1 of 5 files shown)


templates/tag-index/tag-index.js
```diff
@@ -157,4 +159,16 @@ export async function loadEager(document) {
 
 export function loadLazy() {
   renderArticles(getArticles());
+  adsenseFunc('tag', 'create');
+}
+
+export async function loadDelayed() {
+  const pageTag = await getTagForUrl();
```
ramboz, line 166: I'm not sure you need to have this in the delayed phase.
It seems to me this could be added at the end of the lazy phase instead as it shouldn't have much impact on the TBT/CLS

ramboz (approved):
Code looks good to me. Just not sure you need to delay the tags ads.

templates/tag-index/tag-index.js
ankyc-phi: that's good to know, i had just separated it out so that the divs are created first and then the ads are displayed - will take it away for another task, and implement this in the next iteration

PR #412, opened 2024-02-28
bugfix: anchor ads for mobile

ankyc-phi:
Branch Commits:

- removed the second/extra cmd push
- slight update to adsense setup function
- doing cmd push inside define slot function
- small refactor to main adsense function

Jira Issue:
https://pethealthinc.atlassian.net/browse/PM-128

Test URLs:
- After:
  - https://bugfix-mobile-anchor-ads--petplace--hlxsites.hlx.live/
  - https://bugfix-mobile-anchor-ads--petplace--hlxsites.hlx.live/?martech=off

Screenshots:

image

### DIFF
--- a/scripts/adsense.js
+++ b/scripts/adsense.js
@@ -67,70 +67,66 @@ const gtagDisplay = (argsArr) => {
 };
 
 const adsenseSetup = (adArgs, catVal) => {
-  let anchorSlot;
   const REFRESH_KEY = 'refresh';
+  const REFRESH_SLOT = 'refreshed_slot';
   const REFRESH_VALUE = 'true';
   const lastItemIndex = adArgs.length - 1;
 
-  window.googletag.cmd.push(() => {
-    // eslint-disable-next-line no-plusplus
-    for (let i = 0; i < lastItemIndex; i++) {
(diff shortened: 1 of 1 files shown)


scripts/adsense.js
```diff
@@ -61,120 +61,111 @@ const adsDivCreator = (adLoc) => {
...
+    renderMarginPercent: 100,
+    mobileScaling: 2.0,
   });
 
+  if (catVal) {
+    const pageType = adArgs[lastItemIndex].split('_')[0];
+    window.googletag.pubads().setTargeting(pageType, catVal);
+  }
+
+  window.googletag.pubads().set('page_url', 'https://www.petplace.com');
```
ramboz, line 126: Will this need to be adjusted for the UK site? if yes, you could already inject the `codeBasePath` here

ramboz (approved):
Looks good :)

scripts/adsense.js
ankyc-phi: good catch, i can update this in #413.....but i think this shouldn't need to be updated, since the website is still "petplace.com" technically - will confirm, and modify in other PR if needed

PR #425, opened 2024-03-13
(PM-129) Adding ByteTag Landing Page

Tom-Scott-Hero:
Fix #PM-129 (https://pethealthinc.atlassian.net/browse/PM-129)

Test URLs:
- Before: https://main--petplace--hlxsites.hlx.live/lost-pet-protection/bytetag
- After:
  - https://pm-129--petplace--hlxsites.hlx.live/lost-pet-protection/bytetag
  - https://pm-129--petplace--hlxsites.hlx.live/lost-pet-protection/bytetag?martech=off


blocks/bytetag-buy/bytetag-buy.js
```diff
@@ -0,0 +1,14 @@
...
+  const container = block.parentElement.parentElement;
+  let {
+    dataset: { bytetagId, bytetagToken },
+  } = container;
+  if (bytetagId && bytetagToken) {
+    if (bytetagId.indexOf('tel:') !== -1) {
+      bytetagId = bytetagId.replace('tel:', '');
+    }
+    loadShopifyBuy(bytetagId, block, bytetagToken);
+  }
```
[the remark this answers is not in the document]
Tom-Scott-Hero: i did try this, but oddly it removes both bytetag embed components.

blocks/columns/columns.css
```diff
@@ -47,6 +47,128 @@
   padding: 0 .5rem;
 }
 
+.columns.cards > div > div {
```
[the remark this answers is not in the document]
Tom-Scott-Hero: Good question, i don't believe this type of layout will be used anywhere outside of the bytetag landing page which is brand specific and has design elements which are different than the typical column layout used across petplace.com

scripts/third-party.js
```diff
@@ -34,3 +34,288 @@ export async function loadLazy() {
...
+    var script = document.createElement('script');
+    script.async = true;
+    script.src = scriptURL;
+    (
+      document.getElementsByTagName('head')[0] ||
+      document.getElementsByTagName('body')[0]
+    ).appendChild(script);
+    script.onload = ShopifyBuyInit;
+  }
+  if (window.ShopifyBuy) {
```
ramboz, line 311: In my tests, I'm seeing the last `else` called multiple times, and we have several calls to `ShopifyBuyInit` and the shopify script is added multiple times to the DOM.

Screenshot 2024-03-13 at 14 52 08

Screenshot 2024-03-13 at 14 52 22

With the script creating TBT for the page, and also taking close to 1s to load on mobile, I'd spend some time optimizing this further.

Screenshot 2024-03-13 at 14 53 36

Tom-Scott-Hero: I agree, unfortunately this embed code was provided by a 3rd party and i had to modify it heavily to make it authorable and functional on the page.  I'll try to streamline this, however i'm hearing they're wanting to go live with this code tomorrow.

ramboz: My suggestion would be to create a single promise that awaits all the way until `ShopifyBuy.UI.onReady(client)`.
You can then use the awaited promise on every call to `createComponent` with the proper `id`

Something along the lines of:
```js
let shopifyReadyPromise;
async function loadShopifyBuy() {
  ...
  if (!shopifyReadyPromise) {
    shopifyLoadedPromise = new Promise((resolve) => {
      const script = document.createElement('script');
      script.async = true;
      script.src = scriptURL;
      document.head.append(script);
      script.onload = () => {
        const client = ShopifyBuy.buildClient({
          domain: 'bytetag.myshopify.com',
          storefrontAccessToken: accessToken,
        });
        ShopifyBuy.UI.onReady(client).then(resolve);
      };
    });
  }
  shopifyReadyPromise.then((ui) => ui.createComponent(…));
}

```

blocks/columns/columns.css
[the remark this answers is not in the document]
ramboz: I'd probably add a comment then so the devs remember this next year when the look at that source code. Makes it easier to avoid regressions down the road if someone starts modifying the wrong stuff 😉

scripts/third-party.js
ramboz: >  i'm hearing they're wanting to go live with this code tomorrow.

I'll let the business decide here… but as it is, the PR would not pass our quality gates seeing the current performance impact and UX. And in my experience, if we let this pass, you likely won't get much bandwidth in the near future to revisit

Tom-Scott-Hero: Agreed, this project has been fairly time constrained. I believe the performance impacts should be limited to just this landing page? Or are you seeing this across other pages?

ramboz: No, it's just that landing page. Essentially the impact of the custom fonts + the shopify scripts

ankyc-phi (approved):
LGTM

PR #432, opened 2024-03-18
PM-15 - GenAI blocks and Search

Tom-Scott-Hero:
Adding Gen AI Functionality and Blocks

Fix #PM-15

Test URLs:
- Before: https://main--petplace--hlxsites.hlx.page/article/cats/pet-care/what-you-need-to-know-before-hiring-pet-sitter
- After:
  - https://pm-15--petplace--hlxsites.hlx.page/discovery
  - https://pm-15--petplace--hlxsites.hlx.page/article/cats/pet-care/what-you-need-to-know-before-hiring-pet-sitter


templates/gen-ai/gen-ai.js
```diff
@@ -0,0 +1,478 @@
...
+  });
+}
+
+const createStreamingSearchCard = (resultsBlock) => {
+  const card = document.createElement('div');
+  card.className = 'search-card';
+  card.classList.add('response-animation');
+  card.innerHTML = `<div class="search-card-container"><div class="search-card-warning"><p>${GENAI_SEARCH_WARNING}</p></div><article></article><div class="slideshow"></div></div>`;
+
+  resultsBlock.innerHTML = card.outerHTML;
```
ramboz, line 226: The `outerHTML` api is usually slow, so you might as well directly all the HTML here, of swap to full DOM APIs and use `append(card)` instead.

templates/gen-ai/gen-ai.js
```diff
@@ -0,0 +1,478 @@
...
+        $slidesContainer.appendChild(linkContainer);
+        slides.push(linkContainer);
+      });
+
+      slides[0].setAttribute('active', true);
+
+      $slideShowContainer.prepend($slidesContainer);
+
+      const $sliderPrev = document.createElement('div');
+      $sliderPrev.innerHTML = '&#60;';
```
ramboz, line 317: I would do those via CSS and `::before` styles

ramboz (commented):
I would suggest extracting all static strins to the placeholders. this will make this easier to localize down the road.

I'd also be tempted to avoid refreshing the page /discovery when a question is submitted, and rather just inline the answer. Plus I'd probably add a default message inviting the user to type a question when they initially reach that page instead of having it just empty

blocks/header/header.css
```diff
@@ -47,6 +47,254 @@
...
+}
+
+body.gen-ai .genai-search-container .default-content-wrapper {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  gap: 20px;
+  margin-bottom: 10px;
+}
+body.gen-ai #search-main .genai-search-container .default-content-wrapper {
```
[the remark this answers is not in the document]
Tom-Scott-Hero: I believe a lot of this code was coming from the original idea of replacing homepage search with gen-ai search.I'll remove this for now

ankyc-phi (approved):
looks good, some css linting errors can be tackled in a future PR after release

ankyc-phi:
@ramboz the css linting errors are held together by "stylelint-ignore" for now, and will be updated at a later time to allow for this to be released

PR #447, opened 2024-03-22
[PPAD-169] [PPAD-170] Landing Page - Mobile and Tablet views

carolinesampaio-herodigital:
[PPAD-169](https://pethealthinc.atlassian.net/browse/PPAD-169) [QA] - Landing Page - Tablet View - The design does not match Figma
[PPAD-170](https://pethealthinc.atlassian.net/browse/PPAD-170) [QA] - Landing Page - Mobile View - The design does not match Figma

Test URLs:
- Before: https://main--petplace--hlxsites.hlx.page/pet-adoption/
- After:
http://localhost:3000/pet-adoption/

Questions: 
1. "The size of the cards should be rendered and match the design in tablet view, Width: 250 px, Height: 385 px"
I fixed the width, but want to confirm first if we should fix the height of the card?

2." PetPlace: More than Just An Adoption Site, title, font size: 49.5 px Description, Font Size: 24 px, Font Type: Inter"
Do we have this font family Inter? It seems to be a bug on Figma.

### DIFF
--- a/blocks/adopt-search/adopt-search.css
+++ b/blocks/adopt-search/adopt-search.css
@@ -231,6 +231,7 @@
   width: 100%;
   min-width: 300px;
   font-size: 18px;
+  font-family: "Libre Franklin";
 
   @media (min-width: 768px) {
     margin-left: auto;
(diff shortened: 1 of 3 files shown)


raymond-p-sanford (approved):
@carolinesampaio-herodigital you are correct, we shouldn't be adjusting the card height, just the width - the height is dependent on the pet images that are displayed

PR #460, opened 2024-04-03
[PPAD-308] Account - Details - Remove search alerts when disabling e-mail consent

carolinesampaio-herodigital:
Screenshot 2024-04-02 at 22 45 32

Screenshot 2024-04-02 at 22 45 45

Test URLs:
http://localhost:3000/pet-adoption/account


raymond-p-sanford (changes requested):
Please update so that user preferences are not saved if they click cancel from within the modal

PR #470, opened 2024-04-10
Adding API error logging for adoption user creation

raymond-p-sanford:
Adding logging for adoption user creation api
Adding additional retry for when user creation api fails

Isolating this update from other PR to expedite getting this logging into production

Test URLs:
- Before: https://www.petplace.com/
- After:
  - https://hero-develop--petplace--hlxsites.hlx.page/
  - https://hero-develop--petplace--hlxsites.hlx.page/?martech=off

### DIFF
--- a/scripts/lib/msal/msal-authentication.js
+++ b/scripts/lib/msal/msal-authentication.js
@@ -5,6 +5,7 @@ import { loginRequest, logoutRequest, tokenRequest, changePwdRequest, msalConfig
 import { isMobile } from '../../scripts.js';
 import endPoints from '../../../variables/endpoints.js';
 import { pushToDataLayer } from '../../utils/helpers.js';
+import { captureError } from '../../scripts.js';
 
 let cachedMsalInstance = null;
 let cachedMsalChangePwdInstance = null;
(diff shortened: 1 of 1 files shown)


scripts/lib/msal/msal-authentication.js
```diff
@@ -298,23 +299,55 @@ function handleResponse(response, customCallback, featureName = 'PetPlace (Gener
...
-        } else {
-            pushToDataLayer({
-                event: 'login',
+          // invoke custom callback if one was provided
+          if (customCallback) {
+            customCallback(response);
+          }
+        })
+        .catch((error) => {
+          captureError('account creation', error);
```
ramboz, line 308: I'd be tempted to extract the whole `fetch(…).then(…).catch(…)` in a dedicated function that you'd call twice so you avoid code duplication. You'd only need an error callback that you set to the method itself on the first call and leave empty on the second call:
```js
 // register the user
 registerUser(response, featureName, success, () => {
   // retry on failure
   setTimeout(() => {
     registerUser(response, featureName, success, () => {});
   }, /* some timeout, like wait 300ms */)
 });
```

ramboz (approved):
Looks good, but I think we could avoid code duplication with minor refactoring and save ourselves a few bytes on each request

PR #481, opened 2024-04-17
[PPAD-416] [QA] - No message displayed when you don't have any results on search page

carolinesampaio-herodigital:
Updated no results page content.

Screenshot 2024-04-17 at 13 29 56

Test URLs:
  - http://localhost:3000/pet-adoption/search?milesRadius=undefined&filterGender=&filterAge=&filterAnimalType=Dog&filterBreed=AFFENPINSCHER&zipPostal=33470&filterSize= 

**Note: The updated label was not changed on placeholders.xlsx yet to avoid possible changes on prod**


blocks/adopt-search-results/adopt-search-results.js
```diff
@@ -189,7 +219,8 @@ async function callAnimalList() {
         }
         const paginationBlock = document.querySelector('.pagination');
         paginationBlock.classList.add('hide');
-        resultsContainer.innerHTML = noResults;
+        resultsContainer.classList.add('no-results');
```
raymond-p-sanford, line 222: @caroline-sampaio we need a corresponding remove 'no-results' if the next search does have results.  Currently doing a search with results after doing one without results, keeps this class which restricts the width of the content

raymond-p-sanford (changes requested):

blocks/adopt-search-results/adopt-search-results.js
carolinesampaio-herodigital: @raymond-p-sanford thanks for pointing that. This is now fixed.

raymond-p-sanford (approved):

PR #502, opened 2024-05-25
PPAD-244: keep filters on pet type change tt

toviastorresherodigital:
Fix #PPAD-244

Test URLs:
http://localhost:3000/pet-adoption/search?milesRadius=10&filterGender=M&filterAge=&filterAnimalType=Dog&filterBreed=&filterShelter=&zipPostal=19121&filterSize=


raymond-p-sanford (changes requested):
@toviastorresherodigital Your update to not reset the filters when changing the animal type is working as expected, however changing the animal type should still trigger a search as soon as the animal type is changed as it currently is doing on production

raymond-p-sanford (approved):

PR #504, opened 2024-05-27
[PPAD-411]: Disable save changes after submit survey tt

toviastorresherodigital:
Disable save changes button after submission returned with 204, and enable the button when user triggers a change on any select input

Fix #PPAD-411

Test URLs:
(http://localhost:3000/pet-adoption/survey?animalType=Dog)


blocks/pet-survey/pet-survey.js
```diff
@@ -576,10 +580,9 @@ export default async function decorate(block) {
         if (accountWrapper) {
```
toviastorresherodigital, line 580: accountWrapper was coming as null @raymond-p-sanford

toviastorresherodigital: what is that one for?

raymond-p-sanford (approved):

PR #527, opened 2024-06-14
PM#344 - Homepage - P3 Aggregator Widget

sagar-pethealth:

Fix #344

Test URLs:
- Before: https://main--petplace--hlxsites.hlx.live/
- After:
  - https://pm-344-aggrator-widget--petplace--hlxsites.hlx.live/
  - https://pm-344-aggrator-widget--petplace--hlxsites.hlx.live/?martech=off

### DIFF
--- a/templates/home-page/home-page.css
+++ b/templates/home-page/home-page.css
@@ -165,6 +165,10 @@ main a:hover {
   text-align: center;
 }
 
+.aggregator-tool {
+  height: 500px;
+}
+
 @media (max-width: 768px) {
   .home-banner-bg {
     margin-bottom: 6rem !important;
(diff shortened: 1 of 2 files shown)


ramboz:
@sagar-pethealth The changes in the code introduce a big TBT spike, and hence impact the overall page performance negatively.
If you take a closer look at the PSI report, you'll see that for the TBT, you have:
Screenshot 2024-06-21 at 11 59 16

Essentially you have:
- 1.5s TBT from the iframe loading
- +350ms from the gtm it is loading
- +180ms from MS Clarity it is loading

I would try to use lazy loading on the iframe as a first step: https://developer.mozilla.org/en-US/docs/Web/Performance/Lazy_loading#images_and_iframes
And if that is not enough, I'd consider loading it when the user scrolls only by using an intersection observer on the parent section.

sagar-pethealth:
@ramboz Thanks!!  The intersection observer on the parent section worked.

ramboz (approved):

ankyc-phi:
@ramboz @sagar-pethealth heads up, we need to make this scalable since blog page will also have the widget - instead of having it in 3 different template files, we can do it through 1 block and load it into whichever page we need

PR #528, opened 2024-06-14
React test POC

darlantc:
This one has the objective to create the POC for the React project running on a specific route of the project.
In this case I created a document `react-test.docx` with the react block.

Please note that this code is not final and the route will not be used in production.

https://github.com/hlxsites/petplace/assets/5101009/03ca6c98-9d14-4dc4-9829-633b53be7530

Test URLs:
- Before: https://dev--petplace--hlxsites.hlx.page/
- After:
  - https://80090-react-test-poc--petplace--hlxsites.hlx.page/
  - https://80090-react-test-poc--petplace--hlxsites.hlx.page/?martech=off


react-client/src/main.tsx
```diff
@@ -0,0 +1,10 @@
+import React from "react";
+import ReactDOM from "react-dom/client";
+import App from "./App.tsx";
+import "./index.css";
+
+ReactDOM.createRoot(document.getElementById("react-root")!).render(
```
ramboz, line 6: Is the trailing `!` intentional?
```suggestion
ReactDOM.createRoot(document.getElementById("react-root")).render(
```

scripts/scripts.js
```diff
@@ -476,6 +476,15 @@ function fixHyperLinks(main) {
     });
 }
 
+async function buildReactBlock(main) {
+  if (!main.innerText.includes('– react –')) return;
```
ramboz, line 480: I'd recommend using the official block notation in the document instead and creating a plain table with the 1st cell set to "react". This will automatically load and execute any logic in `/blocks/react/react.js`, and you avoid having to parse actual strings.

| React |
|-|
| |

you can also use it to potentially pass down configs into the react block:


| React (style-variant-1, style-variant-2) |
|-|
| some key | some value |
| some other key | some other value |

and leverage the `readBlockConfig` in `lib-franklin.js` to retrieve this as a JSON map.
So you'd have a markup like:
```html
<div class="block react style-variant-1 style-variant-2"></div>
```
a config like:
```json
{
  "some-key": "some-value",
  "some-other-key": "some-other-value"
}
```
And you could isolate the code in a bloc decoration like:
```js
import { readBlockConfig } from '../scripts/lib-franklin.js'
export default function decorate(block) {
  const config = readBlockConfig(block);
  block.innerHTML = '';
  import('./react-index-COxdwC34.js');
}
```

The block would likely also be the sole content in the document, so you could also update the `LCP_BLOCKS` in `scripts.js` and add `react` to it, so we await the rendering in the eager phase before we show the page (limits CLS).

Another approach could also be to leverage the templates construct and have one or more dedicated react templates to render the page.

You'd annotate the document with:
| Metadata |
|-|
| Template | react |

And then define the logic in the template's JS file.

Both approaches have the benefit of:
1. isolating the code from the main `scripts.js`
2. only executing when really needed, and not on every single page. The `main.innerText.includes(…)` is going to be pretty expensive on large pages

ramboz (commented):
Looks overall good. I'd suggest refactoring and isolating how the react logic is initialized so it doesn't leak on every single page, but is rather only executed when needed via some metadata or block opt-in

search-index.db
ramboz, line 1: This file should not be modified (it's updated via async workflows and never directly in PRs).
I'd make sure to properly rebase against `main` or manually revert the changes to be safe.

ramboz (approved):
Code is looking good to me.

A few overall dev/maintenance workflow questions come to mind though:
- as we inject more dependencies that are likely to have security issues, should we maybe also look at introducing renovate (or similar) to automate dependency management?
- we introduce a whole new building system as part of this PR, should we also maybe properly update the GH workflows to automatically run the (post)build steps?

search-index.db
darlantc: I don't know why git says there is a diff since it came from a rebase I did from main, not something I changed manually on my branch.

I tried, but there is nothing to revert 👀

PR #534, opened 2024-06-17
[SITES-22691] Track user region source in RUM data

rachel-kittens:

Fix #

Test URLs:
- After:
  - https://22691-track-region-in-rum--petplace--rachel-kittens.hlx.page/
  - https://22691-track-region-in-rum--petplace--rachel-kittens.hlx.page/?martech=off

### DIFF
--- a/scripts/scripts.js
+++ b/scripts/scripts.js
@@ -94,6 +94,15 @@ window.hlx.plugins.add('rum-conversion', {
 // eslint-disable-next-line prefer-rest-params
 function gtag() { window.dataLayer.push(arguments); }
 
+/**
+ * Logs the information about an error encountered by the site.
+ * @param {string} source Description of the source that generated the error.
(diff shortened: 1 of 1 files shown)


scripts/scripts.js
```diff
@@ -993,6 +993,8 @@ async function loadLazy(doc) {
   sampleRUM('lazy');
   sampleRUM.observe(main.querySelectorAll('div[data-block-name]'));
   sampleRUM.observe(main.querySelectorAll('picture > img'));
+  const region = await getRegion();
```
ekremney, line 996: here, you could use the `promise.then` instead of `await` in order not to block the page from rendering while getting the region

ramboz: Good catch!

rachel-kittens:
Modify scripts.js to track user region source in RUM data

PR #536, opened 2024-06-19
title, author and hero design modified according to figma

ankyc-phi:
Jira Ticket: https://pethealthinc.atlassian.net/browse/PM-310

Test URLs:
- After:
  - https://feat-new-article-template-title-hero-author--petplace--hlxsites.hlx.live/
  - https://feat-new-article-template-title-hero-author--petplace--hlxsites.hlx.live/?martech=off

### DIFF
--- a/templates/article-page/article-page.js
+++ b/templates/article-page/article-page.js
@@ -1,43 +1,5 @@
-import {
-  buildBlock,
-  decorateBlock,
-  loadBlock,
-  getMetadata,
-  toClassName,
-} from '../../scripts/lib-franklin.js';
-import {
-  createBreadCrumbs,
-  getCategories,
-} from '../../scripts/scripts.js';
-import { pushToDataLayer } from '../../scripts/utils/helpers.js';
+import { buildBlock, getMetadata } from '../../scripts/lib-franklin.js';
 
-const GENAI_TOOLTIP = 'Try our AI powered discovery tool and get all your questions answered';
-
-export async function getCategoryByKey(key, value) {
-  const categories = await getCategories();
-  return categories.find((c) => c[key].toLowerCase() === value.toLowerCase());
-}
-
-function createAutoBlockSection(main, blockName, gridName) {
-  const gridNameValue = gridName || blockName;
-  const section = document.createElement('div');
-  section.classList.add('article-template-autoblock', `article-template-grid-${gridNameValue}`);
-
-  main.append(section);
-  return section;
-}
-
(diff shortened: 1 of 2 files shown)


templates/article-page/article-page.js
```diff
@@ -56,191 +18,47 @@ function createTableOfContents(main) {
...
-  breadcrumb.style.visibility = 'hidden';
-  breadCrumbs.append(breadcrumb);
-  decorateBlock(breadcrumb);
-  await loadBlock(breadcrumb);
-  breadcrumb.style.visibility = '';
-
-  const { adsenseFunc } = await import('../../scripts/adsense.js');
-  adsenseFunc('article', 'create');
-
-  const createGenAISearchCTA = () => {
```
ramboz, line 207: Is it really intended that the GenAI search gets removed in the new design?

templates/article-page/article-page.js
```diff
@@ -56,191 +18,47 @@ function createTableOfContents(main) {
...
+
+  const articleTitle = main.querySelectorAll('h1')[0];
+  const authorDiv = main.querySelector('.article-author-container');
+  authorDiv.classList.remove('section');
+  const heroImgContainer = main.querySelectorAll('p')[0];
+  heroImgContainer.classList.add('hero-pic-div');
+  heroTitleSection.append(articleTitle);
+  heroTitleSection.append(authorDiv);
+  heroTitleSection.append(heroImgContainer);
+  main.prepend(heroTitleSection);
```
ramboz, line 59: All these DOM manipulations in the lazy phase are likely to trigger CLS.
I wonder if you could maybe move those up in the eager phase instead, and work directly on the undecorated markup. You can take inspiration from the auto blocking logic in scripts.js that decorates the hero block for instance

ramboz (commented):
Hard for me to evaluate without access to the mocks this is meant to implement, so I'll trust you on the CSS side.

For the JS, see my comment

templates/article-page/article-page.js
ankyc-phi: it's only getting removed for the time being, i will be adding it back in once the template is fully ready - i have the original file saved in my local, plus it will be on main

templates/article-page/article-page.js
ankyc-phi: i tried moving up to the eager phase, but there's errors in the console for not being able to find the classlist of the containers/divs that i'm targetting since they're not loaded yet

PR #542, opened 2024-06-24
80518 design system base components

mat3usps:
Test URLs:
- After:
  - https://80518-design-system-base-components--petplace--hlxsites.hlx.live/
  - https://80518-design-system-base-components--petplace--hlxsites.hlx.live/?martech=off


darlantc:
Please take a look into the lint, it failed on the styles.css file changes

react-client/tailwind.config.js
```diff
@@ -4,13 +4,52 @@ module.exports = {
   content: ["./src/**/*.{js,jsx,ts,tsx}"],
   theme: {
     colors: {
-      "highlight-color": "var(--highlight-color)",
-      "primary-color": "var(--cta-button-color)",
+      "purple-500": "var(--purple-500: #302280)",
```
darlantc, line 7: The way you defined the var usage doesn't work. Please follow this pattern:
```"purple-500": "var(--purple-500)",```

File: react-client/src/components/switch/Switch.tsx [hunk 2]
```diff
@@ -0,0 +1,30 @@
+import classNames from "classnames";
```
darlantc, line 1: You forgot to add classnames as a dependency on the project.

However, I prefer to use the new [clsx](https://www.npmjs.com/package/clsx) package, which is better optimized.

react-client/src/components/switch/styles.css
```diff
@@ -0,0 +1,43 @@
+@import '@radix-ui/colors/black-alpha.css';
```
darlantc, line 1: I don't see the need to use a separated css file for such implementation.
Prefer using Tailwind classes in the component itself.

react-client/src/components/icon/Icon.tsx
```diff
@@ -0,0 +1,30 @@
+import type { FontAwesomeIconProps } from "@fortawesome/react-fontawesome";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
```
darlantc, line 2: I don't expect to use font-awesome in this project.
Please use .svg files exported from Figma.

PR #546, opened 2024-06-27
Revert "PM#389 - PPL: Replace ads.txt File"

ankyc-phi:
Reverts hlxsites/petplace #545


ankyc-phi:
reverting this commit to get it tested first

PR #554, opened 2024-07-01
80089 Prevent errors on account activation

yuliaribeiro:
This PR has the objective to prevent errors when request from API the picture and name of an animal.
Notice that by default only the three first animals will have their pictures rendered.

Here bellow I present the html behaving when:

1.  Has only one animal on list and has no picture on the DB, but the animalType is dog
Screen Shot 2024-07-01 at 4 06 37 PM

2.  Has only one animal on list and no picture on the DB, but the animalType is cat
Screen Shot 2024-07-01 at 4 07 39 PM

3.  Has more than one animal on list, and the second one has no picture on the DB and no animalType was provided.
Screen Shot 2024-07-01 at 4 02 08 PM

4.  Has more than one animal on list, and the last one has no picture
Screen Shot 2024-07-01 at 4 05 23 PM

**Please ignore the URLs bellow.**
Test URLs:
- After:
  - https://80089-prevent-errors--petplace--hlxsites.hlx.live/
  - https://80089-prevent-errors--petplace--hlxsites.hlx.live/?martech=off


auth-pages/account-activation.js
```diff
@@ -42,7 +52,7 @@ function renderPetsPictures(petsList) {
 
 		const imageTag = document.createElement("img");
 		imageTag.src = src;
-
+		imageTag.alt = "Pet image";
```
darlantc, line 53: Let's try to make the alt dynamic, based on the pet name to differentiate the images for blind people.

PR #573, opened 2024-07-08
article cta box for article page

ankyc-phi:
Branch Commits:
- article cta created
- adding article cta to the page

Jira Ticket:
https://pethealthinc.atlassian.net/browse/PM-313

Test URLs:
- After:
  - https://feature-new-article-template-article-cta--petplace--hlxsites.hlx.live/
  - https://feature-new-article-template-article-cta--petplace--hlxsites.hlx.live/?martech=off


blocks/article-cta/article-cta.js
```diff
@@ -0,0 +1,19 @@
...
+  const details = document.createElement('span');
+  const button = document.createElement('button');
+
+  heading.textContent = 'Pet Insurance: A Pet Care Must Have';
+  details.textContent = 'Compare the top pet insurance plans all in one place to find the one that works best for you and your pet';
+
+  img.classList.add('article-cta-img');
+  img.src = '/icons/compare-plans.svg';
+  button.textContent = 'Compare Plans';
+  button.href = '/pet-insurance';
```
ramboz, line 13: This feels like a pretty "static" block. I wonder if it would not be easier to create a fragment that you load here, so content is easier to change. You could then just call the `createTemplateBlock` method with the right `elems` config with the fragment URL so it inlines it automatically for you

ramboz (approved):

blocks/article-cta/article-cta.js
ankyc-phi: added a comment to the file for fragment, i'll work on it as phase 2

PR #577, opened 2024-07-09
page alignment for grid in mobile and desktop

ankyc-phi:
Branch Commits:
- adding gen ai back in
- adding attribute setting back
- setting up content alignment
- css changes for grid layout
- changes for divs attached to ads
- breadcrumbs added to page
- minor changes to element positioning

Jira Ticket:
https://pethealthinc.atlassian.net/browse/PM-343

Test URLs:
- After:
  - https://feat-new-article-page-alignment--petplace--hlxsites.hlx.live/
  - https://feat-new-article-page-alignment--petplace--hlxsites.hlx.live/?martech=off

### DIFF
--- a/blocks/genai-search/genai-search.css
+++ b/blocks/genai-search/genai-search.css
@@ -35,8 +35,8 @@
   }
 }
 .section.genai-search-container:not(.centered) {
-  border-top: 1px solid var(--color-gray-lighter-alt);
-  padding-top: 1rem;
+  border-top: 2px solid var(--border-color);
+  padding: 0;
 
   @media (min-width: 1200px) {
     max-width: 1344px;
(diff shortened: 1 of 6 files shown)


blocks/genai-search/genai-search.css
```diff
@@ -1172,6 +1162,10 @@ body:not(.gen-ai) .genai-search.large .search-box-wrapper #clearButton {
   border-color: var(--background-color) transparent transparent transparent;
 }
 
+.article-page .header-search {
+  width: auto !important;
```
ramboz, line 1166: We typically avoid using `!important` as this usually prevents proper overrides in larger media queries. Have you validated that this worked across all sizes? Also would there be a more specific selector you could use to avoid the important and just use regular specificity?

ramboz (approved):

blocks/genai-search/genai-search.css
ankyc-phi: it's actually not needed anymore, good spot - i'll take it out, thank you!!

PR #586, opened 2024-07-10
80899 add pet profile info

yuliaribeiro:
This PR is responsible to render all information when clicking to view an animal profile

https://github.com/hlxsites/petplace/assets/89787597/4423e4ff-7d9d-4940-8a15-c737738426b4

Test URLs:
- After:
  - https://80899-add-pet-profile-info--petplace--hlxsites.hlx.page/account/my-pets
  - https://80899-add-pet-profile-info--petplace--hlxsites.hlx.page/account/my-pets/?martech=off


react-client/src/components/Pet/useCardBase.tsx
mat3usps, line 1: I don't see a real need for the creation of theses constants

react-client/src/routes/pet-profile/PetProfileIndex.tsx
```diff
@@ -13,7 +16,9 @@ export const PetProfileIndex = () => {
         primaryElement={renderActionsButton()}
         shouldRenderBackButton
       />
```
mat3usps, line 18: If you are already testing the existence of petInfo, why do it again when passing the props  to the children components? It seems redundant. I'd suggest destructuring petInfo to keep it less verbose.

react-client/src/components/Pet/useCardBase.tsx
yuliaribeiro: I thought that the team agreed yesterday that this could be used in another places, and we just needed to move it to the right place as Darlan suggested. Also this file belongs to the last PR that was approved. If we definitely need to remove from the project please let me know.

PR #622, opened 2024-07-30
81752 table component

mat3usps:
image

Test URLs:
- After:
  - https://81752-table-component--petplace--hlxsites.hlx.live/
  - https://81752-table-component--petplace--hlxsites.hlx.live/?martech=off


react-client/src/components/design-system/text/Title.tsx
```diff
@@ -1,14 +1,24 @@
 import { classNames } from "~/util/styleUtil";
 
-type TitleProps = {
+type TitleVariableProps = {
+  align?: "center" | "left" | "right" | "justify";
+  level?: "h1" | "h2" | "h3" | "h4" | "h5" | "h6";
+  fullWidth?: boolean;
```
yuliaribeiro, line 6: I have similar opinion about the fullWidth.. Using it we can accidentally change the line-height and letter-spacing, the text or title shouldn't worry about width

react-client/src/routes/playground/playgrounds/ConditionalPlayground.tsx
yuliaribeiro, line 1: I'm really not fan of this component..
Let's discuss with the team if we really need to implement in the project..
So, here in the playground you are using 3 text, but the playground is rendering 2.
I needed to open the code to understand the logic why is showing and in what conditions.
My suggestion when writing playgrounds would be think about someone that needs to test the element without opening code to understand the logic.. If the team decides to maintain this Conditional, you can add buttons or select with options containing the "condition" when text will appear and when not.

react-client/src/components/design-system/table/Table.tsx
yuliaribeiro, line 1: I saw that you implemented on the Table the functionality to select items inside it, although, I didn't find in the project any table with this feature. Maybe this went unnoticed by me, but If we won't use, then I don't see as benefit to write here. This is very useful I have no doubt, but I would prefer to keep the project as clean as possible only with features that we should rely on.
So, with this in mind, I leave everything related to the Table component to @darlantc review. As a tech leader he is the best choice to evaluate what we should implement or not for this particular project.

yuliaribeiro (changes requested):
Mateus good job :)
I did a generic review, but when I was reviewing the table a saw an implementation that I wasn't sure.
I left there a comment so @darlantc can evaluate better how we should proceed.

react-client/src/components/design-system/table/Table.tsx
```diff
@@ -0,0 +1,264 @@
...
+    }
+    return columnsList;
+  }
+
+  function renderNoResultsView() {
+    if (!isEmptyState || isLoading) {
+      return null;
+    }
+
+    const message = noResultsMessage ?? "Nenhum resultado encontrado para";
```
yuliaribeiro, line 251: My suggestions for the message here are: 
1. talk with the design team and check the appropriated copy for this case.
2. use English

react-client/src/components/design-system/table/Table.tsx
```diff
@@ -0,0 +1,264 @@
...
+    return (
+      <p>
```
yuliaribeiro, line 253: We already have the Text component, instead of using p tag with strong, I'd recommend to use the Text and TextSpan. We have similar code in the project so you can inspire.

react-client/src/components/design-system/table/Table.tsx
```diff
@@ -0,0 +1,264 @@
...
+          </div>
+        </Card>
+      </div>
+    );
+  }
+
+  function renderAsTable() {
+    const tableColumns = getColumns();
+    return (
+      <table aria-label="table" className="min-w-full table-auto">
```
yuliaribeiro, line 108: I would improve this aria-label to be more assertive, remember that screen readers can count on this attribute, imagine the scenario where we have 2 tables..
To fix this you can pass aria-label as prop.

react-client/src/components/design-system/table/Table.tsx
```diff
@@ -0,0 +1,264 @@
...
+        <Conditional
+          when={tableColumns.filter(({ label }) => !!label).length > 0}
```
yuliaribeiro, line 110: nit: I believe that you can replace this 
`tableColumns.filter(({ label }) => !!label).length > 0` 
to this:
`tableColumns.some(({ label }) => !!label)`

As I understood, the first one filters out columns without a truthy label and then checks if there are any columns left by comparing the length to 0.
Using the some you can directly checks if there is at least one column with a truthy label without needing to filter and then check the length.

react-client/src/components/design-system/table/Table.tsx
```diff
@@ -0,0 +1,264 @@
...
+              <tr
+                className={classNames("hover table-row w-fit")}
+                data-testid="table_tbody_tr"
+                key={`row-${row.key}`}
+              >
+                <Conditional when={isSelectable}>
+                  <td className={classNames(cellClassName({ isEven }))}>
+                    <Checkbox
+                      id={row.key}
+                      label="Selecionar item"
```
yuliaribeiro, line 144: I suggest to change this label to English

react-client/src/components/design-system/table/Table.tsx
```diff
@@ -0,0 +1,264 @@
...
+        </tbody>
+      </table>
+    );
+  }
+
+  function getColumns(): TableColumn[] {
+    const actionsColumn: TableColumn = {
+      align: "right",
+      key: "actionsColumn",
+      label: "Ações",
```
yuliaribeiro, line 187: Same here... use English

react-client/src/components/design-system/table/Table.tsx
```diff
@@ -0,0 +1,264 @@
...
+      const isAllSelected = totalSelectedItems === rows.length;
+
+      columnsList.push({
+        key: "inputCheckbox",
+        label: (
+          <Checkbox
+            checked={isAllSelected}
+            hideLabel
+            id="select-all"
+            label="Selecionar todos os itens"
```
yuliaribeiro, line 203: Same here.. use English

react-client/src/components/design-system/table/Table.tsx
```diff
@@ -0,0 +1,264 @@
...
+import { plural } from "~/util/stringUtils";
+import { Button, Card, Checkbox, Loading } from "~/components/design-system";
+import { Conditional } from "../conditional/Conditional";
+import type { IconKeys } from "../icon/Icon";
+import { TableHeader } from "./TableHeader";
+import { TableRowActions } from "./TableRowActions";
+import { classNames } from "~/util/styleUtil";
+import { useCellBase } from "./useCellBase";
+
+export interface TableColumn {
```
yuliaribeiro, line 11: you are exporting a lot of interfaces without necessity, my suggestion is to create a file for TableTypes. then you can set every type there and share just what you need. We are already doing similar thing on the project, you can use the `DialogBaseTypes` or the `FormTypes` to inspire.

I personally avoid to use Interfaces because they can be merged together. If you declare the same interface multiple times, TypeScript will automatically merge them and this can be a little issue for maintenance if they are declared in different files. But this is just a particular thing why I prefere types instead of interfaces

react-client/src/components/design-system/table/Table.tsx
```diff
@@ -0,0 +1,264 @@
...
+  align?: "left" | "center" | "right";
+  className?: string;
+  isSortable?: boolean;
+  key: string;
+  label?: ReactNode;
+  width?: number | string;
+}
+
+export interface TableRow<T> {
```
yuliaribeiro, line 20: nit: this exporting is being used only on the `Table.test.tsx` my suggestion is to avoid as max as possible the imports created because of the tests, ideally they need to rely just what is going on inside the component and we shouldn't create or change anything for tests. That's the reason why we extensively use ComponentProps

react-client/src/components/design-system/table/Table.test.tsx
```diff
@@ -0,0 +1,604 @@
...
+const AGE_COLUMN = /^age$/i;
+const EMAIL_COLUMN = /^email$/i;
+
+describe("<Table />", () => {
+  const testRows = createRows(6);
+
+  describe("with isLoading={true}", () => {
+    it("should display LoadingView", () => {
+      const { container } = getRenderer({
+        isLoading: true,
```
yuliaribeiro, line 37: nit: instead of writing isLoading: true every time, you can define on the top of this describe (after line 34) the `const isLoading = true`
and on the getRenderer you can use something like:
`getRenderer({ isLoading })`
If you want to you can apply the same when isLoading = false
this is not an important suggestion and applying or not won't affect the code.

react-client/src/components/design-system/table/Table.test.tsx
```diff
@@ -0,0 +1,604 @@
...
+  getByLabelText,
+  getByTestId,
+  getByText,
+  queryByLabelText,
+  queryByRole,
+  queryByTestId,
+  queryByText,
+} = screen;
+
+const NO_RESULTS_MESSAGE = "Nenhum resultado encontrado para";
```
yuliaribeiro, line 22: you need to change the const: `NO_RESULTS_MESSAGE`, `SELECT_ALL_CHECKBOX` and `SELECT_ITEM_CHECKBOX` that are in portuguese

react-client/src/components/design-system/table/Table.test.tsx
```diff
@@ -0,0 +1,604 @@
...
+  });
+
+  describe("with isLoading={false}", () => {
+    it("should NOT display LoadingView", () => {
+      getRenderer({
+        isLoading: false,
+        rows: createRows(2),
+      });
+      expect(queryByTestId("LoadingView")).not.toBeInTheDocument();
+    });
```
yuliaribeiro, line 66: this test is false positive, this testId does not exist on the project, it is working only because query can return undefined. fix this assertion to find the loading correctly. We already have some parts in the project that are doing similar things, you can use them to inspire. This change is important.

After this I would also suggest to change the text because we don't have a LoadingView component, but we have the Loading component, and this fix can be applied on other parts that you are calling the "LoadingView"

react-client/src/components/design-system/table/Table.test.tsx
```diff
@@ -0,0 +1,604 @@
...
+        expect(getAllByRole("columnheader")).toHaveLength(3);
+      });
+
+      it("should NOT display Columns", () => {
+        getRenderer({
+          columns: [],
+          rows: createRows(4),
+        });
+        expect(queryByRole("columnheader")).not.toBeInTheDocument();
+        expect(queryByTestId("table_thead_th")).not.toBeInTheDocument();
```
yuliaribeiro, line 107: This test is also a false positive, just because you are using the query selector.
The test id `table_thead_th` does not exist on the Table component.
Change the assertion and try to use getByRole associated with name props to find an element.

react-client/src/components/design-system/table/Table.test.tsx
```diff
@@ -0,0 +1,604 @@
...
+      it("should NOT display Columns with empty or undefined label", () => {
+        getRenderer({
+          columns: [
+            { isSortable: true, key: "name", label: undefined },
+            { isSortable: true, key: "email", label: "" },
+          ],
+          rows: createRows(4),
+        });
+        expect(queryByRole("columnheader")).not.toBeInTheDocument();
+        expect(queryByTestId("table_thead_th")).not.toBeInTheDocument();
```
yuliaribeiro, line 119: Same problem of the false positive here.. change the assertion

react-client/src/components/design-system/table/Table.test.tsx
yuliaribeiro, line 107: I suggest to avoid using the `queryByTestId` or `getByTestId`, you should be able to find the elements using the role associated with the prop `name` or combining them with the `within` method.
Using the data-testid in _reusable components_ is considered hardcoding and might to lead to a false positive, that's why relying on labels and roles is always a  better approach.

react-client/src/components/design-system/table/Table.test.tsx
```diff
@@ -0,0 +1,604 @@
...
+      });
+
+      describe("rowActions", () => {
+        it("should NOT display 'actionColumn' when no rowActions are provided", () => {
+          getRenderer({
+            rowActions: [],
+            rows: createRows(5),
+          });
+          expect(
+            queryByRole("columnheader", { name: "Ações" })
```
yuliaribeiro, line 333: use English instead of Portuguese

react-client/src/components/design-system/table/Table.test.tsx
```diff
@@ -0,0 +1,604 @@
...
+          ).not.toBeInTheDocument();
+        });
+
+        it("should display 'actionColumn' when at least one rowActions are provided", () => {
+          getRenderer({
+            rowActions: [{ icon: "check", id: "add-note", label: "Add Note" }],
+            rows: createRows(5),
+          });
+          expect(
+            getByRole("columnheader", { name: "Ações" })
```
yuliaribeiro, line 343: same issue about Portuguese here

react-client/src/components/design-system/table/TableRowActions.tsx
yuliaribeiro, line 1: Is missing the test for this component

react-client/src/components/design-system/table/Table.tsx
```diff
@@ -0,0 +1,264 @@
...
+      {wrapper ? wrapper(renderTableContent()) : renderTableContent()}
+      </Conditional>
+    </>
+  );
+
+  function renderTableContent() {
+    return (
+      <div className="w-auto overflow-hidden py-large">
+        <Card>
+          <div className="scrolling-touch overflow-x-auto">
```
yuliaribeiro, line 97: I'm not an expert in tailwind, but maybe overflow should be `overflow-auto` to include the ability to scroll tables vertically as well.. If this is being applied in another place, please ignore this comment.

react-client/src/components/design-system/text/Title.tsx
```diff
@@ -1,14 +1,24 @@
 import { classNames } from "~/util/styleUtil";
 
-type TitleProps = {
+type TitleVariableProps = {
+  align?: "center" | "left" | "right" | "justify";
```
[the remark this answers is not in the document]
mat3usps: There are designs models that need the align prop for the title, like the one below.

image

react-client/src/components/design-system/text/Title.tsx
mat3usps: There are designs that need the width of title to be full to work with centered titles like the one below.

image

PR #638, opened 2024-08-03
81843 - Added Pet edit form

darlantc:
screenshot 2024-08-03 at 19 35 42@2x

Test URLs:
- Before: https://dev--petplace--hlxsites.hlx.page/
- After:
  - https://pet-edit-form--petplace--hlxsites.hlx.live/
  - https://pet-edit-form--petplace--hlxsites.hlx.live/?martech=off


react-client/src/components/design-system/form/Select.tsx
yuliaribeiro, line 1: The accessibility when using keyboard when select is open is not working. It's not possible to navigate in the items or select them. My suggestion would be to control the focus when select is open, so user can use the keyboard.

react-client/src/routes/my-pets/:petId/edit/PetEditIndex.tsx
```diff
@@ -1,7 +1,44 @@
...
+            console.log("onChange values", props);
+          }}
+          onSubmit={({ event, values }) => {
+            event.preventDefault();
+
+            console.log("onSubmit values", values);
+          }}
+          schema={editPetProfileFormSchema}
+          variables={{
+            // This could come from an API request, for example
```
yuliaribeiro, line 26: nit: I would use the mockUseApi and place there all things that need to be mocked in the project.  Because when we remove the mock we can also see better all places that will break and the fix will be easier. By using mock it's self explanatory as well, we don't need to signalize every time that some piece of code could come from an API. This is just a personal way how to organize things, feel free to use or not.

react-client/src/routes/my-pets/:petId/form/petForm.tsx
yuliaribeiro, line 1: nit: I understand what you did here, although the file could get very long and the maintenance could be difficult if we end up adding more forms related to pets or even fields. In this case, I would place the variables in a exclusive file, then separate the edit and the add forms in different files, because they have their own responsibilities..

yuliaribeiro (commented):
very clever this form builder. great job.

mat3usps (approved):

PR #655, opened 2024-08-08
fix: GenAI Trigger Words: CTA Card Loading Issues

hero-dokane:
**Issues:** 
- By default after a query is submitted, we will show a ‘insurance placeholder cta card’. 
- If there is cta card content specific to the search query returned we will replace the ‘insurance placeholder cta card’ with that content.
- Move logic for creation of the CTA Card from type: streaming to type: end 
- Update logic to only output the first response.actions item as a cta card

Fix: [PM-455](https://pethealthinc.atlassian.net/browse/PM-455)

Test URLs:
- Before: https://main--petplace--hlxsites.hlx.page/discovery
- After:
  - https://bugfix-genai-trigger-words-cta-card--petplace--hlxsites.hlx.page/discovery
  - https://bugfix-genai-trigger-words-cta-card--petplace--hlxsites.hlx.page/discovery/?martech=off


ankyc-phi (approved):
lgtm

PR #691, opened 2024-08-21
Enabled folder mapping to fix react client redirect

darlantc:
Test URLs:
- After:
  - https://fix-404-redirect--petplace--hlxsites.hlx.live/
  - https://fix-404-redirect--petplace--hlxsites.hlx.live/?martech=off

### DIFF
--- a/404.html
+++ b/404.html
@@ -26,8 +26,10 @@
       window.location.replace('/rss.xml');
     }
 
-    // Redirect React app
-    if (window.location.pathname.includes('/account')) {
+    // Redirect to React client app
+    const isProdSite = window.location.origin.includes('petplace.com');
+    // It is disabled for prod site while we are not ready to release it
+    if (!isProdSite && window.location.pathname.includes('/account')) {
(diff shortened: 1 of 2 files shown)


ramboz (commented):
Trying to wrap my head around the use case here.
Am I right to assume we'd essentially have customers navigating to:
```
https://www.petplace.com/account/some-sub-page
```
which doesn't exist in EDS.
So we'd trigger the 404 page, which would then redirect again to:
```
https://www.petplace.com/account?redirectFrom=/some-sub-page
```
?

If yes, how would customers obtain `https://www.petplace.com/account/some-sub-page` in the first place?
Wouldn't it be best to reflect a valid URL instead, and work with:
```
https://www.petplace.com/account#/some-sub-page
```
with some form of hash-based routing instead?

If the goal is just to render a react-managed 404 instead, then I'd probably work with a "fragments" approach, and load a react view inside the 404 page so you can customize the 404 according to your needs under the /account section

PR #717, opened 2024-08-30
Implemented account activation page

darlantc:
I added the `account-activation` template, which redirects the users to the correct Azure policy.

https://github.com/user-attachments/assets/8854b781-4be5-4720-bb77-3ceb5a608050

Test URLs:
- After:
  - https://82820-account-activation--petplace--hlxsites.hlx.live/
  - https://82820-account-activation--petplace--hlxsites.hlx.live/?martech=off

### DIFF
--- a/templates/account-activation/account-activation.js
+++ b/templates/account-activation/account-activation.js
@@ -1,0 +1,45 @@
+import { redeemTokenRequest } from '../../scripts/lib/msal/default-msal-config.js';
+import { getDefaultMsalInstance } from '../../scripts/lib/msal/msal-instance.js';
+
+function parseJwt(token) {
+  if (!token) return null;
+
+  const base64Url = token.split('.')[1];
+  const base64 = base64Url.replace(/-/g, '+').replace(/_/g, '/');
+  const jsonPayload = decodeURIComponent(window.atob(base64).split('').map((c) => `%${(`00${c.charCodeAt(0).toString(16)}`).slice(-2)}`).join(''));
+
+  return JSON.parse(jsonPayload);
+}
+
+export async function loadEager(document) {
+  const searchParams = new URLSearchParams(window.location.search);
+  const token = searchParams.get('token');
+  const claim = parseJwt(token);
+
+  if (!token || !claim) {
+    const main = document.querySelector('main');
+    main.innerHTML = `
+        <div>
+            <h1>Invalid token</h1>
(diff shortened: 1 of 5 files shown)


templates/account-activation/account-activation.js
```diff
@@ -0,0 +1,45 @@
...
+  const base64 = base64Url.replace(/-/g, '+').replace(/_/g, '/');
+  const jsonPayload = decodeURIComponent(window.atob(base64).split('').map((c) => `%${(`00${c.charCodeAt(0).toString(16)}`).slice(-2)}`).join(''));
+
+  return JSON.parse(jsonPayload);
+}
+
+export async function loadEager(document) {
+  const searchParams = new URLSearchParams(window.location.search);
+  const token = searchParams.get('token');
+  const claim = parseJwt(token);
```
ramboz, line 17: this could potentially throw an error (from the `JSON.parse`) if we have a corrupt token, so might be good to wrap in an try/catch and have the  "invalid token" message shown in that case as well.

tools/importer/import.js
```diff
@@ -487,6 +487,13 @@ export default {
       case 'article-page':
         main = transformArticlePage(document);
         break;
+      case 'account-activation':
```
ramboz, line 490: Did you actually use the content importer for this PR? Do you need these changes?

darlantc: Oh, I wanted to use it but could not figure out how it works. I ended up hiding those things with CSS.
That page doesn't need the header & footer.

templates/account-activation/account-activation.js
darlantc: Nice catch, thank you!

tools/importer/import.js
ramboz: The importer is actually a tool we use to initially create the website based on an existing setup on a different system. We essentially crawl the whole sitemap and for each page we parse the DOM and create the matching documents in Sharepoint. It's not really a tool we use once in production and we just kept this around in case there was further content that needed to be migrated to EDS.

If you are interested, you can check https://www.aem.live/developer/importer for more details, but you can otherwise just revert those changes as they won't have any impact on the running site

darlantc: Oh, got it. Thank you!

ramboz (approved):

PR #789, opened 2024-09-28
84005 renew api

yuliaribeiro:
Test URLs:
- After:
  - https://84005-renew-api--petplace--hlxsites.hlx.live/
  - https://84005-renew-api--petplace--hlxsites.hlx.live/?martech=off


yuliaribeiro:
The use case built here should be applied on the pet profile

